PR #17, opened 2020-04-14
Setup proper logging

Sammers21:
The only way to debug JRuby code is via logs, that is why the changes are introduced:

* sl4j-simple has been replaced with dedicated logback configuration which prints debug logs during tests


g4s8:
@Sammers21 can we use log4j as in parent pom?

Sammers21:
@g4s8, yes, done

0crat:
Job `gh:artipie/gem-adapter#17` is not assigned, can't get performer

PR #19, opened 2020-04-16
Pass Storage instance to SubmitGem slice

Sammers21:
Since SubmitGem need Storage access, we need to pass it via constructor


0crat:
Job `gh:artipie/gem-adapter#19` is not assigned, can't get performer

PR #28, opened 2020-04-23
Add test for gem push

Sammers21:
The test ensures that gem-adapter is compatible with `gem push` operation.

The test is disabled on windows since testcontainers does not support Windows

### DIFF
--- a/src/test/java/com/artipie/gem/GemCliITCase.java
+++ b/src/test/java/com/artipie/gem/GemCliITCase.java
@@ -24,47 +24,67 @@
 package com.artipie.gem;
 
 import com.artipie.asto.memory.InMemoryStorage;
-import com.artipie.http.rs.RsStatus;
 import com.artipie.vertx.VertxSliceServer;
 import io.vertx.reactivex.core.Vertx;
-import io.vertx.reactivex.core.buffer.Buffer;
-import io.vertx.reactivex.ext.web.client.WebClient;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import org.hamcrest.MatcherAssert;
-import org.hamcrest.core.IsEqual;
+import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledIfSystemProperty;
+import org.testcontainers.Testcontainers;
+import org.testcontainers.containers.Container;
+import org.testcontainers.containers.GenericContainer;
 
 /**
- * A test for gem submit operation.
+ * A test which ensures {@code gem} console tool compatibility with the adapter.
  *
  * @since 0.2
+ * @checkstyle StringLiteralsConcatenationCheck (500 lines)
  */
-public class SubmitGemITCase {
+@SuppressWarnings("PMD.SystemPrintln")
+@DisabledIfSystemProperty(named = "os.name", matches = "Windows.*")
+public class GemCliITCase {
 
     @Test
-    public void submitNotImplemented() throws IOException {
+    public void gemPushWorks() throws IOException, InterruptedException {
         final Vertx vertx = Vertx.vertx();
         final VertxSliceServer server = new VertxSliceServer(
             vertx,
             new GemSlice(new InMemoryStorage(), vertx.fileSystem())
         );
-        final WebClient web = WebClient.create(vertx);
         final int port = server.start();
-        final byte[] gem = Files.readAllBytes(
-            Paths.get("./src/test/resources/builder-3.2.4.gem")
+        final String host = String.format("http://host.testcontainers.internal:%d", port);
+        Testcontainers.exposeHostPorts(port);
+        final RubyContainer ruby = new RubyContainer()
+            .withCommand("tail", "-f", "/dev/null")
+            .withWorkingDirectory("/home/")
+            .withFileSystemBind("./src/test/resources", "/home");
+        ruby.start();
+        final Container.ExecResult push = ruby.execInContainer(
+            "/bin/bash",
+            "-c",
+            String.format("GEM_HOST_API_KEY=123 gem push builder-3.2.4.gem --host %s", host)
         );
-        final int code = web.post(port, "localhost", "/api/v1/gems")
-            .rxSendBuffer(Buffer.buffer(gem))
-            .blockingGet()
-            .statusCode();
+        System.out.println(push.getStdout());
+        System.err.println(push.getStderr());
         MatcherAssert.assertThat(
-            code,
-            new IsEqual<>(Integer.parseInt(RsStatus.OK.code()))
+            String.format("'gem push builder-3.2.4.gem --host %s' failed with non-zero code", host),
+            push.getExitCode(),
+            Matchers.equalTo(0)
         );
-        web.close();
+        ruby.stop();
         server.close();
         vertx.close();
+    }
+
+    /**
+     * Inner subclass to instantiate Ruby container.
+     *
+     * @since 0.1
+     */
+    private static class RubyContainer extends GenericContainer<RubyContainer> {
+        RubyContainer() {
+            super("ruby:2.7");
+        }
     }
 }
--- a/src/test/java/com/artipie/gem/GemCliITCase.java
+++ b/src/test/java/com/artipie/gem/GemCliITCase.java
@@ -25,6 +25,7 @@ package com.artipie.gem;
 
 import com.artipie.asto.memory.InMemoryStorage;
 import com.artipie.vertx.VertxSliceServer;
+import com.jcabi.log.Logger;
 import io.vertx.reactivex.core.Vertx;
 import java.io.IOException;
 import org.hamcrest.MatcherAssert;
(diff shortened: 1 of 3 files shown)


src/test/java/com/artipie/gem/Base.java
```diff
@@ -0,0 +1,48 @@
...
+ * @checkstyle NonStaticMethodCheck (500 lines)
+ */
+public class Base {
+
+    /**
+     * Find a random port.
+     * @return A random port.
+     * @throws IOException if fails.
+     */
+    protected int randomPort() throws IOException {
```
g4s8, line 43: @Sammers21 you can use `reserver-network-port` standard Maven plugin instead of this class:
```xml
          <plugin>
            <groupId>org.codehaus.mojo</groupId>
            <artifactId>build-helper-maven-plugin</artifactId>
            <version>3.0.0</version>
            <executions>
              <execution>
                <id>reserver-random-port</id>
                <goals>
                  <goal>reserve-network-port</goal>
                </goals>
                <phase>pre-integration-test</phase>
                <configuration>
                  <portNames>
                    <portName>vertx.port</portName>
                  </portNames>
                  <minPortNumber>49152</minPortNumber>
                  <maxPortNumber>65535</maxPortNumber>
                  <randomPort>true</randomPort>
                </configuration>
              </execution>
            </executions>
          </plugin>
          <plugin>
            <artifactId>maven-failsafe-plugin</artifactId>
            <configuration>
              <systemPropertyVariables combine.children="append">
                <vertx.port>${vertx.port}</vertx.port>
              </systemPropertyVariables>
            </configuration>
          </plugin>
```

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -0,0 +1,75 @@
...
+            new GemSlice(new InMemoryStorage(), vertx.fileSystem()),
+            port
+        );
+        server.start();
+        MatcherAssert.assertThat(
+            String.format("'gem push builder-3.2.4.gem --host %s' failed with non-zero code", host),
+            new ProcessBuilder()
+                .directory(new File("./src/test/resources"))
+                .command(
+                    "gem",
```
g4s8, line 58: @Sammers21 it's a not good idea to require a developer to install Ruby just to build Java project. I think Docker with Ruby using [Testcontainers](https://www.testcontainers.org/) would be better here.

Sammers21: So you will require a developer to install a docker instead?

g4s8: @Sammers21 yes

src/test/java/com/artipie/gem/Base.java
Sammers21: I think to find a random port in Java is better: it is better maintainable, takes fewer lines of code, more elegant, not tied with maven

g4s8: @Sammers21 the reason why I don't like `new Socket` implementation is that it uses side effect of socket instance: the purpose of using `new Socket` is to perform network operations via this socket. Also, this class has some design problems, so I think it would be harder to fix it, then use maven plugin. Another problem I see, is that we're introducing some functionality which already provided by the build and test system.

Sammers21: This plugin is doing just the same, but implicitly.

Sammers21: And we already are using the random port approach everywhere, why should we use the plugin in this case?

g4s8: @Sammers21 I think maven plugin is more stable, also it supports multiple random ports. But you are right, let's use this plugin in all other artipie modules.

Sammers21: > @Sammers21 I think maven plugin is more stable

How can you prove it? 

I have also checked and found out that, this approach is incompatible with default IDE test runners. It means that you would not be able to debug/run tests from IntelliJ Idea for example without an extra hassle. Sounds like a very strange decision.

g4s8: @Sammers21 just check source code of this plugin, it has much more conditional checks and exception handling

g4s8:
@Sammers21 it's really hard to read the history of this PR. Please try to remove commits "for debugging on CI machine" and keep only relevant commits with self-describing messages. Also, don't use this draft PR feature - it doesn't make any sense here, when changes are not ready for review why do you submit PR with it?

scrot-2020-04-28-16:53:49

src/test/java/com/artipie/gem/Base.java
Sammers21: > @Sammers21 just check source code of this plugin, it has much more conditional checks and exception handling

It does not matter how many checks is there, how can you prove that one approach is more "stable" than another. The current code never fails, what makes you think that it is not stable. Moreover, integration tests for Artipie are not military piece software. There is no requirement to be ULTRA stable. 

@g4s8 and you have not responded to the broken IDE support complaint. Are you okay with sacrificing it in favor of non-confirmed stability?

g4s8: @Sammers21 never fails - how can you prove it? Maven plugin ensures that acquire port is available twice, your code can generate random port which will be taken by another process in the middle of method calls.
IDE is a not a good place for running integration test, it's ok to run unit tests in IDE, but integration tests should be run by correctly configured software. I don't see any problem with no running integration tests in IDE.
The main question is why do we need to write custom code for testing that's already implemented by build tools.

Sammers21: > @g4s8, IDE is a not a good place for running integration test, it's ok to run unit tests in IDE, but integration tests should be run by correctly configured software. I don't see any problem with no running integration tests in IDE.

It is subjective. Ask any member of your team, any of them will be against the sacrifice. It is just common sense. To sacrifice any significant part of developer experience chasing some phantom improvements - a bad idea in general.

Sammers21:
@g4s8, review it again please

src/test/java/com/artipie/gem/Base.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+import java.net.ServerSocket;
+
+/**
+ * A base class with common utilities.
+ *
+ * @since 0.2
+ * @checkstyle NonStaticMethodCheck (500 lines)
+ */
+public class Base {
```
g4s8: @Sammers21 now you can delete this class

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -0,0 +1,90 @@
...
+import org.testcontainers.containers.GenericContainer;
+
+/**
+ * A test which ensures {@code gem} console tool compatibility with the adapter.
+ *
+ * @since 0.2
+ * @checkstyle StringLiteralsConcatenationCheck (500 lines)
+ */
+@SuppressWarnings("PMD.SystemPrintln")
+@DisabledIfSystemProperty(named = "os.name", matches = "Windows.*")
```
g4s8, line 45: @Sammers21 why it's disabled on Windows?

src/test/resources/log4j.properties
```diff
@@ -1,7 +1,8 @@
...
+log4j.rootLogger=INFO, CONSOLE
 
 log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
 log4j.appender.CONSOLE.layout=com.jcabi.log.MulticolorLayout
 log4j.appender.CONSOLE.layout.ConversionPattern=[%color{%p}] %t %c: %m%n
 
-log4j.logger.com.artipie=DEBUG
\ No newline at end of file
+log4j.logger.com.artipie=DEBUG
+log4j.logger.org.testcontainers=DEBUG
```
g4s8, line 8: @Sammers21 do we really need debug logs of testcontainers?

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -0,0 +1,90 @@
...
+            .withCommand("tail", "-f", "/dev/null")
+            .withWorkingDirectory("/home/")
+            .withFileSystemBind("./src/test/resources", "/home");
+        ruby.start();
+        final Container.ExecResult push = ruby.execInContainer(
+            "/bin/bash",
+            "-c",
+            String.format("GEM_HOST_API_KEY=123 gem push builder-3.2.4.gem --host %s", host)
+        );
+        System.out.println(push.getStdout());
```
g4s8, line 68: @Sammers21 can we use logger here to control the level via `log4j.properties`?

src/test/java/com/artipie/gem/Base.java
Sammers21: @g4s8, switched to Vert.x server provided random port.

Sammers21:
@g4s8 , review the pull request one again, please.

src/test/resources/log4j.properties
Sammers21: Yes, I find them useful.

src/test/java/com/artipie/gem/GemCliITCase.java
Sammers21: Because Windows is not supported:
`[WARN] main org.testcontainers.dockerclient.DockerClientProviderStrategy: windows is currently not supported`

Sammers21:
@g4s8 , review the Pull Request once again please.

0crat:
Job `gh:artipie/gem-adapter#28` is not assigned, can't get performer

PR #77, opened 2020-09-28
Use random name for temp dir

### LINKED ISSUE #76: Use different dirs for different gem adapter repos
Now all the repos stored in a single temp dir. This way would not allow to use more then 1 instance of gem adapter in an instance of artipie.

Sammers21:
Test is not required for this PR.

Instead of "temp-gem-index", a name with random uuid is used as a temp dir.

Closes #76

### DIFF
--- a/src/main/java/com/artipie/gem/GemSlice.java
+++ b/src/main/java/com/artipie/gem/GemSlice.java
@@ -77,7 +77,8 @@ public final class GemSlice extends Slice.Wrap {
         this(storage,
             JavaEmbedUtils.initialize(new ArrayList<>(0)),
             Permissions.FREE,
-            (login, pwd) -> Optional.of(new Authentication.User("anonymous"))
+            (login, pwd) -> Optional.of(new Authentication.User("anonymous")),
+            "default"
         );
     }
 
@@ -88,11 +89,13 @@ public final class GemSlice extends Slice.Wrap {
      * @param runtime The Jruby runtime.
      * @param permissions The permissions.
      * @param auth The auth.
+     * @param repo The repo identifier.
      */
     public GemSlice(final Storage storage,
         final Ruby runtime,
         final Permissions permissions,
-        final Authentication auth) {
+        final Authentication auth,
+        final String repo) {
         super(
             new SliceRoute(
                 new RtRulePath(
@@ -105,7 +108,7 @@ public final class GemSlice extends Slice.Wrap {
                             runtime,
                             "SubmitGem",
                             storage,
-                            GemSlice.tmpDir()
+                            GemSlice.tmpDirForRepo(repo)
                         ),
                         new Permission.ByName(permissions, Action.Standard.WRITE),
                         new GemApiKeyIdentities(auth)
@@ -173,13 +176,14 @@ public final class GemSlice extends Slice.Wrap {
 
     /**
      * Temp dir, which will be removed on jvm exit.
+     * @param repo The repo identifier.
      * @return The dir name.
      */
-    private static String tmpDir() {
+    private static String tmpDirForRepo(final String repo) {
         final String fname = String.join(
             "-",
-            "gem-tmp-index",
-            UUID.randomUUID().toString()
+            repo,
+            "gem-tmp-index"
         );
         Runtime.getRuntime().addShutdownHook(
             new Thread(
--- a/src/main/java/com/artipie/gem/GemSlice.java
+++ b/src/main/java/com/artipie/gem/GemSlice.java
@@ -185,17 +185,6 @@ public final class GemSlice extends Slice.Wrap {
             repo,
             "gem-tmp-index"
         );
-        Runtime.getRuntime().addShutdownHook(
-            new Thread(
-                () -> {
-                    try {
-                        FileUtils.deleteDirectory(new File(fname));
-                    } catch (final IOException exc) {
-                        throw new UncheckedIOException(exc);
-                    }
-                }
-            )
-        );
         return fname;
     }
 }
--- a/src/main/java/com/artipie/gem/GemSlice.java
+++ b/src/main/java/com/artipie/gem/GemSlice.java
@@ -39,11 +39,14 @@ import com.artipie.http.rt.RtRulePath;
 import com.artipie.http.rt.SliceRoute;
 import com.artipie.http.slice.SliceDownload;
 import com.artipie.http.slice.SliceSimple;
+import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Optional;
+import java.util.UUID;
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 import org.jruby.Ruby;
 import org.jruby.RubyRuntimeAdapter;
@@ -98,7 +101,12 @@ public final class GemSlice extends Slice.Wrap {
                         new RtRule.ByPath("/api/v1/gems")
                     ),
                     new SliceAuth(
-                        GemSlice.rubyLookUp("SubmitGem", storage, runtime),
+                        GemSlice.rubyLookUp(
+                            runtime,
+                            "SubmitGem",
+                            storage,
+                            GemSlice.tmpDir()
+                        ),
                         new Permission.ByName(permissions, Action.Standard.WRITE),
                         new GemApiKeyIdentities(auth)
                     )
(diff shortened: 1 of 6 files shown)


olegmoz (changes requested):
@Sammers21 thanks, please see some comments

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -154,11 +163,35 @@ private static Slice rubyLookUp(final String rclass,
...
+            new Thread(
+                () -> {
+                    try {
+                        FileUtils.deleteDirectory(new File(fname));
+                    } catch (final IOException exc) {
+                        throw new UncheckedIOException(exc);
+                    }
+                }
+            )
+        );
```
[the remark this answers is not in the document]
Sammers21: @olegmoz, Bad decision performance wise. Index update would not be incremental.

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -98,7 +101,12 @@ public GemSlice(final Storage storage,
                         new RtRule.ByPath("/api/v1/gems")
                     ),
                     new SliceAuth(
-                        GemSlice.rubyLookUp("SubmitGem", storage, runtime),
+                        GemSlice.rubyLookUp(
+                            runtime,
+                            "SubmitGem",
+                            storage,
+                            GemSlice.tmpDir()
+                        ),
```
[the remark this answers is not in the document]
Sammers21: @olegmoz, this pr is not related to the problem of concurrent submission.

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -154,11 +163,35 @@ private static Slice rubyLookUp(final String rclass,
...
+    /**
+     * Temp dir, which will be removed on jvm exit.
+     * @return The dir name.
+     */
+    private static String tmpDir() {
+        final String fname = String.join(
+            "-",
+            "gem-tmp-index",
+            UUID.randomUUID().toString()
+        );
```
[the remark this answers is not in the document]
Sammers21: @olegmoz, does not work with createTempDir for unknown reasons

src/main/java/com/artipie/gem/GemSlice.java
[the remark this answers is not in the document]
olegmoz: @Sammers21 directory is created on every request this way, so everything will be lost between requests anyway

src/main/java/com/artipie/gem/GemSlice.java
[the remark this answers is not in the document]
olegmoz: @Sammers21 then please add tests to verify problem this PR solves

src/main/java/com/artipie/gem/GemSlice.java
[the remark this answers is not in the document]
olegmoz: @Sammers21 please investigate what is the problem with temp directory, because otherwise we store our files in a place user does not expect to be polluted with temporary data

src/main/java/com/artipie/gem/GemSlice.java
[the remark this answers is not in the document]
Sammers21: @olegmoz, it is hard to create a test.

src/main/java/com/artipie/gem/GemSlice.java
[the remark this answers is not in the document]
Sammers21: @olegmoz, done

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -74,7 +76,8 @@ public GemSlice(final Storage storage) {
         this(storage,
             JavaEmbedUtils.initialize(new ArrayList<>(0)),
             Permissions.FREE,
-            (login, pwd) -> Optional.of(new Authentication.User("anonymous"))
+            (login, pwd) -> Optional.of(new Authentication.User("anonymous")),
+            "default"
```
olegmoz, line 80: @Sammers21 no sure that's a great idea to allow creating `GemSlice` with default repo name. User should specify name explicitly to avoid storing data in same directory

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -98,7 +103,12 @@ public GemSlice(final Storage storage,
                         new RtRule.ByPath("/api/v1/gems")
                     ),
                     new SliceAuth(
-                        GemSlice.rubyLookUp("SubmitGem", storage, runtime),
+                        GemSlice.rubyLookUp(
+                            runtime,
+                            "SubmitGem",
+                            storage,
+                            GemSlice.tmpDirForRepo(repo)
+                        ),
```
olegmoz, line 111: @Sammers21 this code needs to be tested. Otherwise we cannot be sure that changes fix the issue

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -154,11 +165,36 @@ private static Slice rubyLookUp(final String rclass,
...
+            new Thread(
+                () -> {
+                    try {
+                        FileUtils.deleteDirectory(new File(fname));
+                    } catch (final IOException exc) {
+                        throw new UncheckedIOException(exc);
+                    }
+                }
+            )
+        );
```
olegmoz, line 197: @Sammers21 I'd suggest to remove this code for directory cleanup, it does not make any sense to clean the directory on application exit. If it is saved between requests then why not save it between application restarts

olegmoz (changes requested):
@Sammers21 thanks, please see the comments

src/main/java/com/artipie/gem/GemSlice.java
Sammers21: @olegmoz, hard to test

olegmoz (changes requested):
@Sammers21 thanks for changes, please see one comment. Also please note that test for the problem solved is mandatory.

Sammers21:
@olegmoz, review the PR one more time, please.

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -48,22 +54,44 @@
...
  * @checkstyle ExecutableStatementCountCheck (500 lines)
  */
-@SuppressWarnings("PMD.SystemPrintln")
+@SuppressWarnings({"PMD.ExcessiveMethodLength", "PMD.AvoidDuplicateLiterals"})
 @DisabledIfSystemProperty(named = "os.name", matches = "Windows.*")
 public class GemCliITCase {
 
     @Test
-    public void gemPushAndInstallWorks(@TempDir final Path temp, @TempDir final Path mount)
+    public void gemPushAndInstallWorks(@TempDir final Path temp)
```
olegmoz, line 62: @Sammers21 I'd suggest to create separate test method instead of making changes to existing one. Otherwise existing test case becomes over-complicated and it will be hard to detect problems with basic functionality in it

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -48,22 +54,44 @@
  * @checkstyle StringLiteralsConcatenationCheck (500 lines)
  * @checkstyle ExecutableStatementCountCheck (500 lines)
  */
-@SuppressWarnings("PMD.SystemPrintln")
+@SuppressWarnings({"PMD.ExcessiveMethodLength", "PMD.AvoidDuplicateLiterals"})
```
olegmoz, line 57: @Sammers21 this `ExcessiveMethodLength` warning is fair, please do not suppress it. The method in the test is way too big and may be simplified by extracting some duplicated code into methods. For example you may consider extracting code for server staring or pushing gems

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -74,41 +102,73 @@ public void gemPushAndInstallWorks(@TempDir final Path temp, @TempDir final Path
...
+            IllegalStateException.class,
+            () -> this.bash(
+                ruby,
+                String.format("GEM_HOST_API_KEY=%s gem fetch -V builder --source %s", key, hosts)
+            ),
+            String.format(
+                "'gem fetch should fail, since builder was not pushed to the second repo",
+                hostf
+            )
+        );
```
olegmoz, line 168: @Sammers21 are you sure this check does fail when single directory is used for both `GemSlice` instances? Because `builder` was present in index, but the `.gem` file was still missing. So to me it seems that this check does not really tests what we need

olegmoz (changes requested):
@Sammers21 thanks, please see the comments

src/test/java/com/artipie/gem/GemCliITCase.java
Sammers21: @olegmoz, the separated test would look the same as this.

src/test/java/com/artipie/gem/GemCliITCase.java
Sammers21: @olegmoz, it's an integration test, it can be  too big

src/test/java/com/artipie/gem/GemCliITCase.java
Sammers21: @olegmoz, look to the test log. It does.

Sammers21:
@olegmoz, review the PR one more time, please.

src/test/java/com/artipie/gem/GemCliITCase.java
Sammers21: @olegmoz, when the same dir used, the check fail

olegmoz:
@Sammers21 your replies do not really answer my questions. Please make required changes or elaborate your answers

Sammers21:
@olegmoz, Почему не отвечают? Я же на все ответил. Ну пиши пожалуйста на русском если я на английском не понимаю.

Sammers21:
@olegmoz, на какой конкретно я не ответил?

0crat:
Pull request [#77](https://github.com/artipie/gem-adapter/issues/77) was not merged, no payment for `ARC`, see [§28](http://www.zerocracy.com/policy.html#28)

PR #82, opened 2020-11-30
#81 - Support Basic authentication

### LINKED ISSUE #81: Support Basic authentication
It would be good to support `Basic` authentication for Gem to install passing credentials in the url. To do this, you need to modify the method `GemApiKeyIdentities#user`

genryxy:
Closes #81 
Support Basic authentication. Add test to verify it

### DIFF
--- a/src/main/java/com/artipie/gem/GemApiKeyIdentities.java
+++ b/src/main/java/com/artipie/gem/GemApiKeyIdentities.java
@@ -24,7 +24,7 @@
 package com.artipie.gem;
 
 import com.artipie.http.auth.Authentication;
-import com.artipie.http.auth.Identities;
+import com.artipie.http.auth.BasicAuthScheme;
 import com.artipie.http.headers.Authorization;
 import com.artipie.http.rq.RqHeaders;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


src/main/java/com/artipie/gem/GemApiKeyIdentities.java
```diff
@@ -55,6 +55,18 @@ public GemApiKeyIdentities(final Authentication auth) {
...
+                    if (str.startsWith(basic)) {
+                        res = str.substring(basic.length());
+                    } else {
+                        res = str;
+                    }
+                    return res;
+                }
+            )
             .map(Base64Decoded::new)
             .map(dec -> dec.toString().split(":"))
```
olegmoz, line 71: @genryxy I'd suggest to use parsing of header value from `Authorized.Basic` class from HTTP module

src/test/java/com/artipie/gem/AuthTest.java
```diff
@@ -132,4 +139,50 @@ public void notAllowedToInstallsUsersAreRejected() throws IOException {
...
+                Arrays.asList(
+                    new RsHasStatus(RsStatus.UNAUTHORIZED),
+                    new RsHasHeaders(new Header("WWW-Authenticate", "Basic"))
+                )
+            )
+        );
+    }
+
+    @Test
+    public void returnsOk() throws IOException {
```
olegmoz, line 157: @genryxy please describe conditions for successful auth in this test's name

src/test/java/com/artipie/gem/AuthTest.java
```diff
@@ -132,4 +139,50 @@ public void notAllowedToInstallsUsersAreRejected() throws IOException {
...
+        MatcherAssert.assertThat(
+            AuthTest.post(true),
+            new RsHasStatus(RsStatus.OK)
+        );
+    }
+
+    private static Response post(final boolean authorized) throws IOException {
```
olegmoz, line 164: @genryxy I think it should be stated in method name that `POST` is done with `Basic` auth header. We test another auth method in this class as well

olegmoz (changes requested):
@genryxy thanks, please see some comments

genryxy:
@olegmoz thanks, I used `Authorization` class after checking that header value starts with `Basic` because in some cases header value can be represented only by token. This required a version update for http module therefore todos were added.

olegmoz (approved):
@genryxy thanks, merging

PR #84, opened 2020-12-07
#83 - Remove deprecated auth classes

### LINKED ISSUE #83: Remove deprecated auth classes
Now in `GemApiKeyIdentities`, `ApiKeySlice`, `GemSlice` suppress warnings annotations are used for deprecation because `com.artipie.http` module was updated from 0.16.2 to 0.17.1 (see #82). So, these annotations should be removed.

genryxy:
Closes #83 
Removed deprecated auth classes. Used `Authentication.Single` and `Permissions.Single` instead of lambda expressions


src/main/java/com/artipie/gem/GemApiKeyAuth.java
```diff
@@ -0,0 +1,132 @@
...
+import com.artipie.http.rq.RqHeaders;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import org.cactoos.text.Base64Decoded;
+
+/**
+ * {@link AuthScheme} implementation for gem api key decoding.
+ * @since 0.4
```
olegmoz, line 39: @genryxy `@since` tag has wrong value, it should be `0.6`

src/main/java/com/artipie/gem/GemApiKeyAuth.java
```diff
@@ -0,0 +1,132 @@
...
+            )
+            .get();
+    }
+
+    /**
+     * Successful result with authenticated user.
+     *
+     * @since 0.5.4
+     */
+    private static class SuccessByToken implements AuthScheme.Result {
```
olegmoz, line 88: @genryxy I'd suggest to rename the class to just `Success`. We do not have any other `Success` classes in the scope, so no need to specify that this `Success` is "by token"

src/main/java/com/artipie/gem/GemApiKeyAuth.java
```diff
@@ -0,0 +1,132 @@
...
+            return "";
+        }
+    }
+
+    /**
+     * Failed result without authenticated user.
+     *
+     * @since 0.5.4
+     */
+    private static class FailureByToken implements AuthScheme.Result {
```
olegmoz, line 120: @genryxy same concern about naming

olegmoz (changes requested):
@genryxy thanks, please see the comments

src/main/java/com/artipie/gem/GemApiKeyAuth.java
genryxy: @olegmoz this class was renamed from `GemApiKeyIdentities`  to `GemApiKeyAuth` but there were many changes and this class looks like a new one. Do I need to change `@since` tag value in this case?

olegmoz: @genryxy yes you do need to update `@since` tag even if you just renamed the class. It breaks compatibility in case it was used somewhere

genryxy:
@olegmoz thanks, corrected

olegmoz (approved):
@genryxy thanks, merging

Issue #15, opened 2020-04-10
GemSlice.java:46-49: Initialize on first request....


0pdd:
The puzzle `13-4aa56782` from #13 has to be resolved: 

https://github.com/artipie/gem-adapter/blob/26af0b4587ecbbc7702d1c124e8c6bc9cb78cb34/src/main/java/com/artipie/gem/GemSlice.java#L46-L49

The puzzle was created by @Sammers21 on 10-Apr-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/farm/issues):

```
PID: 12@cf6a468ba0d0, thread: pool-792-thread-1
com.zerocracy.radars.github.ThrottledComments[74] java.lang.IllegalStateException: Can't post an identical message to artipie/gem-adapter#15: @g4s8[/z](https://www.0crat.com/u/g4s8) please, pay attention to this issue
```

[0.53.1](https://github.com/zerocracy/farm): Issue: artipie/gem-adapter#15, Action: `opened`

0pdd:
The puzzle `13-4aa56782` has disappeared from the source code, that's why I closed this issue.

PR #92, opened 2021-04-21
Rewrite slice implementation from ruby to java refactoring

### LINKED ISSUE #90: Rewrite slice implementation from ruby to java
Remove existing Ruby files and implement `Slice` on Java using `com.artipie.http` packages, reuse `Gem` SDK implementation for Gem index updates instead of built-in scripts. Keep existing integration tests to verify that new implementation works find as HTTP repository.

milkysoft:
GemSlice calls newly created Slice implementation, which:

1. Saves attached gem file into given storage.
2. Creates metadata for newly saved gem using Gem class.
3. Returns CREATED result.

close #90


g4s8:
@milkysoft please see https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md#pull-request-style

g4s8:
@milkysoft please make sure the CI is passing

src/main/java/com/artipie/gem/SubmitGem.java
```diff
@@ -0,0 +1,89 @@
...
+import com.artipie.http.slice.ContentWithSize;
+import java.nio.ByteBuffer;
+import java.util.Map.Entry;
+import java.util.UUID;
+import org.reactivestreams.Publisher;
+
+/**
+ * A slice, which servers gem packages.
+ *
+ * @since 1.0
```
olenagerasimova, line 42: @milkysoft since tag value is incorrect

src/main/java/com/artipie/gem/SubmitGem.java
```diff
@@ -0,0 +1,89 @@
...
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ * @checkstyle ParameterNumberCheck (500 lines)
```
olenagerasimova, line 44: @milkysoft these suppressions does not seem necessary

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -39,20 +39,11 @@
...
-import org.jruby.RubyRuntimeAdapter;
-import org.jruby.javasupport.JavaEmbedUtils;
 
 /**
  * A slice, which servers gem packages.
  *
- * @todo #13:30min Initialize on first request.
  *  Currently, Ruby runtime initialization and Slice evaluation is happening during the GemSlice
  *  construction. Instead, the Ruby runtime initialization and Slice evaluation should happen
  *  on first request.
```
olenagerasimova, line 49: @milkysoft this part of javadoc was part of the [puzzle](https://github.com/artipie/gem-adapter/issues/15) and does not seem valid now. Please, correct it to correspond current implementation or remove it.

olenagerasimova (changes requested):
@milkysoft sorry for the delay, please, check some minor comments

Sammers21:
This change is a step backwards in a direction of making the Ruby adapter independent from the local file system. The right way is to change `AstoIndexer.rb` in a way so it uses asto storage only without filesystem at all. It does not make any sense to replace the current implementation with a simplistic java wrapper that does not even support incremental update and generate index every time from the scratch. And I'd also warn you that the implementation you are trying to switch will not work as good or might be not working at all.

src/main/java/com/artipie/gem/SubmitGem.java
```diff
@@ -0,0 +1,89 @@
...
+     * @return The Slice
+     */
+    public Response response(final String line, final Iterable<Entry<String, String>> headers,
+        final Publisher<ByteBuffer> body) {
+        return new AsyncResponse(
+            this.storage.save(
+                new Key.From(
+                    UUID.randomUUID().toString().replace("-", "")
+                ), new ContentWithSize(body, headers)
+            ).thenCompose(none -> this.gem.batchUpdate(new Key.From("TestKey")))
```
g4s8, line 85: `TestKey` seems to be a typo

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -70,8 +70,21 @@ public void gemPushAndInstallWorks(@TempDir final Path temp, @TempDir final Path
...
+        final Path builder = mount.resolve("gems/builder-3.2.4.gem");
         Files.copy(Paths.get("./src/test/resources/builder-3.2.4.gem"), bgem);
         Files.copy(Paths.get("./src/test/resources/rails-6.0.2.2.gem"), rgem);
+        Files.copy(Paths.get("./src/test/resources/test/latest_specs.4.8.gz"), latestspecs);
+        Files.copy(Paths.get("./src/test/resources/test/prerelease_specs.4.8.gz"), prerelease);
+        Files.copy(
+            Paths.get("./src/test/resources/test/quick/Marshal.4.8/builder-3.2.4.gemspec.rz"),
+            buildergem
+        );
+        Files.copy(Paths.get("./src/test/resources/test/gems/builder-3.2.4.gem"), builder);
```
g4s8, line 87: I'd suggest getting rid of one time variables and use string literals instead. Also, accessing `./src/test` path is not a correct way of copying test resources, please use proper classloader methods. And finally, it could be grouped in a loop for better readability.

g4s8 (changes requested):

Sammers21:
@g4s8 , it seems that you are not paying enough attention, but the adapter maximum scale will be 100 artifacts max with unreasonable update time like a few seconds. If you feel like you want to bury this adapter then of course merge this pull request.

src/main/java/com/artipie/gem/SubmitGem.java
milkysoft: @g4s8 , fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,11 +90,11 @@
     /**
      * Batch update Ruby gems for repository.
      *
-     * @param prefix Location of repository
+     * @param gem Location of repository
```
g4s8, line 93: The javadoc is not correct here

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,11 +90,11 @@
...
      * @return Completable action
      */
-    public CompletionStage<Void> batchUpdate(final Key prefix) {
-        final Storage remote = new SubStorage(prefix, this.storage);
+    public CompletionStage<Void> batchUpdate(final Key gem) {
+        final Storage remote = this.storage;
```
g4s8, line 97: This variable assignment is useless (just copy the reference)

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -137,10 +138,23 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+    private static CompletionStage<Void> copyStorage(final Storage src, final Storage dst,
+        final Key gem) {
+        final List<String> vars = new ArrayList<>(7);
+        vars.add("latest_specs.4.8");
+        vars.add("latest_specs.4.8.gz");
+        vars.add("prerelease_specs.4.8");
+        vars.add("prerelease_specs.4.8.gz");
+        vars.add("specs.4.8");
+        vars.add("specs.4.8.gz");
+        if (gem != null) {
```
g4s8, line 153: `gem` we don't perform null-checking for internal code since Checkstyle won't allow returning null or null assignment.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -137,10 +138,23 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+            vars.add(gem.toString());
+        }
         return Single.fromFuture(src.list(Key.ROOT))
+            .filter(pair -> vars.contains(pair))
```
g4s8, line 157: This `filter` method will reject `list` result if vars doesn't contain `pairs`; we need to filter out elements which are not part of both sets, in other words we need an intersection of key list with our metadata and gem:
result = list &cap; (metadata &cup; gem)

src/main/java/com/artipie/gem/SubmitGem.java
```diff
@@ -82,7 +82,7 @@ public Response response(final String line, final Iterable<Entry<String, String>
                 new Key.From(
                     UUID.randomUUID().toString().replace("-", "")
                 ), new ContentWithSize(body, headers)
-            ).thenCompose(none -> this.gem.batchUpdate(new Key.From("TestKey")))
+            ).thenCompose(none -> this.gem.batchUpdate(Key.ROOT))
```
g4s8, line 85: It should be gem key

g4s8 (changes requested):
@milkysoft please see my comments

g4s8:
@milkysoft also, the CI seems to be broken

src/main/java/com/artipie/gem/SubmitGem.java
milkysoft: @olenagerasimova , we agreed the next release will be 1.0, this is the first time this file will be included into release (together with GemIndex.java, Gem.java), thus 1.0 seems to be correct

src/main/java/com/artipie/gem/SubmitGem.java
milkysoft: @olenagerasimova , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/20d38eafac34381139e01f77472be20594c9bddb)

src/main/java/com/artipie/gem/GemSlice.java
milkysoft: @olenagerasimova , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/20d38eafac34381139e01f77472be20594c9bddb)

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8, thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/20d38eafac34381139e01f77472be20594c9bddb)

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/20d38eafac34381139e01f77472be20594c9bddb)

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/20d38eafac34381139e01f77472be20594c9bddb)

src/main/java/com/artipie/gem/SubmitGem.java
milkysoft: @g4s8, thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/20d38eafac34381139e01f77472be20594c9bddb)

src/test/java/com/artipie/gem/GemCliITCase.java
milkysoft: @g4s8 , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/68b72b0e4f107d93f548b14923e9e1d170382d69)

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -137,10 +140,27 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+        vars.add("prerelease_specs.4.8.gz");
+        vars.add("specs.4.8");
+        vars.add("specs.4.8.gz");
+        vars.add(gem.string());
         return Single.fromFuture(src.list(Key.ROOT))
+            .map(
+                list -> list.stream().filter(
+                    key -> {
+                        return vars.contains(key.string())
+                            || key.string().contains("quick/Marshal.4.8");
```
g4s8, line 161: We can add this key to the set

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -137,10 +140,27 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
      * Copy storage from src to dst.
      * @param src Source storage
      * @param dst Destination storage
+     * @param gem Key for gem
      * @return Async result
      */
-    private static CompletionStage<Void> copyStorage(final Storage src, final Storage dst) {
+    private static CompletionStage<Void> copyStorage(final Storage src, final Storage dst,
+        final Key gem) {
+        final Set<String> vars = new HashSet<>();
```
g4s8, line 148: I'd rename this set to self-describing variable. And it could be places in a static field. To simplify constructing you can use `new HasSet<>(Arrays.asList("latest_specs.4.8", "latest_specs.4.8.gz", "etc.."))`

src/main/java/com/artipie/gem/GemSlice.java
```diff
@@ -39,22 +39,12 @@
...
-import org.jruby.RubyRuntimeAdapter;
-import org.jruby.javasupport.JavaEmbedUtils;
 
 /**
  * A slice, which servers gem packages.
  *
- * @todo #13:30min Initialize on first request.
- *  Currently, Ruby runtime initialization and Slice evaluation is happening during the GemSlice
- *  construction. Instead, the Ruby runtime initialization and Slice evaluation should happen
+ *  Ruby runtime initialization and Slice evaluation happens
```
g4s8, line 47: What does this javadoc mean?

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -68,29 +73,39 @@ public void gemPushAndInstallWorks(@TempDir final Path temp, @TempDir final Path
...
-            ),
-            Matchers.equalTo(0)
-        );
-        Files.delete(bgem);
-        Files.delete(rgem);
+        final Set<String> gems = new HashSet<>();
+        gems.add("builder-3.2.4.gem");
+        gems.add("rails-6.0.2.2.gem");
+        final Iterator<String> iter = gems.iterator();
+        while (iter.hasNext()) {
```
g4s8, line 81: There could be simple `for (String resource : gems)`

g4s8 (changes requested):
@milkysoft please check a few comments

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/d6ec4b7678e9b8c28627464fa1d1d60842906fb1)

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/d6ec4b7678e9b8c28627464fa1d1d60842906fb1)

src/test/java/com/artipie/gem/GemCliITCase.java
milkysoft: @g4s8 , thanks, fixed in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/d6ec4b7678e9b8c28627464fa1d1d60842906fb1)

src/main/java/com/artipie/gem/GemSlice.java
milkysoft: @g4s8, I changed to `HTTP Layer` in latest [commit](https://github.com/artipie/gem-adapter/pull/92/commits/a695fa360b5c951ade2ec13e0671e4e609898625)

g4s8 (approved):
@milkysoft thanks

g4s8:
@olenagerasimova please approve if your comments were resolved

olenagerasimova (approved):
@milkysoft thanks

PR #96, opened 2021-06-11
GemInfo interface

### LINKED ISSUE #85: Implement `GemInfo` slice
Let's properly implement `GemInfo` slice provide information by json or yaml from request:
 - [x] HTTP `Slice` implementation for Gem info #94
 - [ ] SDK method for producing Gem info data in valid format

milkysoft:
Closes #85:
 1. Add GemInfo interface
 2. Create GemInfoClass to extract gem information json similar to SubmitGem
 3. Make Gem class use both, GemIndex and GemInfo
 4. Rewrite unit tests to extract Gem Info JSON using Gem SDK


src/main/java/com/artipie/gem/Gem.java
```diff
@@ -54,6 +56,11 @@
  */
 public final class Gem {
 
+    /**
+     * Gem indexer shared instance cache.
+     */
+    private final AtomicReference<GemInfo> infocache;
```
g4s8, line 62: These two atomic-references looks very similar, both interfaces (`GemInfo` and `GemIndex`) could be instantiated simultaneously, since both uses atomic references only for lazy loading of ruby runtime and ruby interpreter

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -74,19 +86,23 @@
      * @param storage Repository storage.
      */
     Gem(final Storage storage) {
-        this(storage, () -> RubyGemIndex.createNew());
+        this(storage, () -> RubyGemIndex.createNew(), () -> RubyObjJson.createNew());
```
g4s8, line 89: Now `RubyGemIndex` and `RubyObjJson` can use same runtime and interpreter instances

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -158,7 +212,7 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
         return Single.fromFuture(src.list(Key.ROOT))
             .map(
                 list -> list.stream().filter(
-                    key -> vars.contains(key.string())
+                    key -> vars.contains(key.string()) || key.string().contains(gem.string())
```
g4s8, line 215: I didn't get this filter expression: why we check if key is in the set of items and then check if gem key converted to string is a substring of key string representation? This code seems to be not obvious.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -187,4 +241,41 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+            )
+        );
+    }
+
+    /**
+     * Find gem in a given path.
+     * @param gem Gem name to get info
+     * @return String full path to gem file
+     */
+    private CompletionStage<Key> getGemFile(final Key gem) {
```
g4s8, line 268: `Key` is not a file, we can't say that this method return a full path then

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -187,4 +241,41 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+        final CompletableFuture<Key> future = new CompletableFuture<>();
+        Single.fromFuture(this.storage.list(Key.ROOT))
+            .map(
+                list -> list.stream().filter(
+                    key -> {
+                        return key.string().contains(gem.string());
```
g4s8, line 274: What this filter supposed to do? E.g. if we have gem called `foo` and we have another gem called `foo-bar`, then you want to find both? I didn't get this

src/main/java/com/artipie/gem/GemInfoClass.java
```diff
@@ -0,0 +1,98 @@
...
+import org.reactivestreams.Publisher;
+
+/**
+ * Returns some basic information about the given gem. GET - /api/v1/gems/[GEM NAME].(json|yaml)
+ * https://guides.rubygems.org/rubygems-org-api/
+ *
+ * @since 0.2
+ */
+@SuppressWarnings({"PMD.UnusedPrivateField", "PMD.SingularField"})
+public final class GemInfoClass implements Slice {
```
g4s8, line 47: I don't this that class is good name for class. And what happens with `GemInfoSlice`?

src/main/java/com/artipie/gem/RubyGemIndex.java
```diff
@@ -62,30 +62,22 @@ public void update(final Path path) {
         this.runtime.eval(
             this.ruby,
             String.format(
-                "Gem::Indexer.new('%s', {build_modern:true}).generate_index",
+                "require 'rubygems/indexer.rb'\n"
```
g4s8, line 65: We can call `require` only once during initialization, like it was implemented already

g4s8 (changes requested):
@milkysoft please check my comment. Also, the commit history looks not accurate, please rebase.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , the concept here is the following. Gem is operating with Keys and Storages. While Ruby itself and in our case - RubyObjJson operates with files. There should be some interface to translate between both. The `gem` in the code above is the name without version. Thus generally speaking you are right, if there are two gems containing the same string, randomly selected the first one will be used to extract info back to the user

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , the concept here is the following. Gem is operating with Keys and Storages. While Ruby itself and in our case - RubyObjJson operates with files. There should be some interface to translate between both.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 the concept here is the following. Usually, the 'gem` in the code above is the gem name, without version and extension. While the file name, which represents by `key` is usually the full gem name, including name, version and extension. While it works in case `info` route, in `dependencies` route case it is required to find the file by pure gem name, which was implemented

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , Single ruby and runtime for booth GemIndexer and GemInfo implemented with latest commit

src/main/java/com/artipie/gem/RubyGemIndex.java
milkysoft: @g4s8 , Require is fixed with latest commit to be called only once for GemIndexer and Geminfo

src/main/java/com/artipie/gem/GemInfoClass.java
milkysoft: @g4s8 , `GemInfoClass` renmaed to `GetGemInfo` similar to `SubmitGem` class with latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , AtomicReference accepts as parameter only one class, thus for each required classes `GemInfo` and `GemIndex` we use separate AtomicReferences.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -74,19 +101,13 @@
...
-
-    /**
-     * New Gem SDK.
-     *
-     * @param storage Repository storage.
-     * @param indexer Gem indexer supplier
-     */
-    Gem(final Storage storage, final Supplier<GemIndex> indexer) {
+        this.runtime = JavaEmbedUtils.newRuntimeAdapter();
+        this.ruby = JavaEmbedUtils.initialize(Collections.emptyList());
```
g4s8, line 105: Please don't perform any heavy operations in constructor

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -31,29 +32,50 @@
...
+import org.jruby.RubyRuntimeAdapter;
+import org.jruby.javasupport.JavaEmbedUtils;
 
 /**
  * An SDK, which servers gem packages.
  * <p>
  * Performes gem index update using specified indexer implementation.
  * </p>
  * @since 1.0
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
```
g4s8, line 60: This suppression is not a good indicator, it means that the class should be decomposed

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -74,19 +101,13 @@
...
-     *
-     * @param storage Repository storage.
-     * @param indexer Gem indexer supplier
-     */
-    Gem(final Storage storage, final Supplier<GemIndex> indexer) {
+        this.runtime = JavaEmbedUtils.newRuntimeAdapter();
+        this.ruby = JavaEmbedUtils.initialize(Collections.emptyList());
         this.storage = storage;
-        this.indexer = indexer;
+        this.indexer = () -> new RubyGemIndex(this.runtime, this.ruby);
```
g4s8, line 107: Creating a new instance is a very fast operation, no need to wrap it with lambdas and atomics. Previously, atomic was used for ruby runtime and interpreter initialization for SDK

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -117,6 +138,44 @@
         ).handle(Gem::removeTempDir);
     }
 
+    /**
+     * Get info Ruby gem.
+     *
+     * @param gem Ruby gem to extract info
+     * @return Completable action
+     */
+    public CompletionStage<JsonObject> getInfo(final Key gem) {
```
g4s8, line 147: No need to use `get` prefix here, `info()` is good enough like a property description.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -117,6 +138,44 @@
...
+            tmpdir -> this.sharedInfo()
+                .thenApply(
+                    rubyjson -> {
+                        final JsonObject obj;
+                        try {
+                            final Key thekey = this.getGemFile(gem).toCompletableFuture().get();
+                            obj = rubyjson.getinfo(
+                                Paths.get(tmpdir.toString(), thekey.string())
+                            );
+                            removeTempDir(tmpdir, null);
```
g4s8, line 169: The exception could be thrown before this line, and temp dir won't be removed in this case. It's a file leakage.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -117,6 +138,44 @@
...
+                        } catch (final InterruptedException | ExecutionException exc) {
```
g4s8, line 170: `InterruptedException` should not cause IO errors, please read how to handle Java interruptions correctly

src/main/java/com/artipie/gem/Gem.java
g4s8: The issue is not solved here: the key may contain the gem name and not being the expected gem key at the same time, so if some key has `abc` in path (e.g. `abc/another-gem`, and gem's name is `abc`, then we may get another gem by mistake. We need to check the name explicitly.

src/main/java/com/artipie/gem/Gem.java
g4s8: When we copy storage objects to file system, then we can operate with `Path`s to avoid mixed abstractions

src/main/java/com/artipie/gem/Gem.java
g4s8: And same here: we need to check it explicitly

src/main/java/com/artipie/gem/GemInfo.java
```diff
@@ -21,158 +21,50 @@
...
-        try {
-            final Optional<String> filename = Files.walk(tmpdir).map(Path::toString)
-                .filter(file -> file.contains(gem) && file.contains(".gem")).findFirst();
-            if (filename.isPresent()) {
-                return filename.get();
-            } else {
-                throw new ArtipieIOException(String.format("Gem %s not found", gem));
+        @Override
+        public JsonObject getinfo(final Path path) {
+            synchronized (this.origin) {
```
g4s8, line 65: If we use share ruby runtime, then we need to synchronize on shared object (the same for `GemInfo` and `GemIndex`). Ideally, we may check if the synchronization is actually needed, since JRuby runtime may have internal and fast synchronization.

g4s8 (changes requested):
@milkysoft please see my comments

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , I would suggest to use `try .. catch... finally` and call `removeTempDir` in `finally { ... }` block

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , agreed, please provide which methods to be moved to another class

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , currently lambdas are used for Supplier, otherwise cache types should be changed. Below, we are considering to remove cache completely

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , here I propose to use `finally` statement to remove temporarily directory in any case

src/main/java/com/artipie/gem/GemInfo.java
milkysoft: @g4s8 , filtering has been adjusted. For cache, any solution is OK for me.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , `getGemFile` functionality is to return the gem file name with version and extension from temporarily directory root folder. It is supposed that all required files are uploaded into Storage root and then they will be copied to temporarily folder root.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , the filtering has been ajusted. Also, to be considered is the following case: when `/api/v1/deendencies` route is called say for **gviz** name, then rubygems returns info for all gems with names including gviz, which is also mentioned in documentation for this route.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -187,4 +264,39 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+     */
+    private CompletionStage<Key> getGemFile(final Key gem) {
+        final CompletableFuture<Key> future = new CompletableFuture<>();
+        Single.fromFuture(this.storage.list(Key.ROOT))
+            .map(
+                list -> list.stream().filter(
+                    key -> key.string().contains(gem.string()) && key.string().endsWith(Gem.GEM_STR)
+                        && !key.string().contains("/")
+                ).limit(1).collect(Collectors.toList()))
+            .flatMapObservable(Observable::fromIterable).forEach(future::complete);
```
g4s8, line 299: `future::complete` - please use `SingleInterop` instead

src/main/java/com/artipie/gem/GemInfo.java
```diff
@@ -21,158 +21,50 @@
...
+public interface GemInfo {
 
     /**
-     * Endpoint path pattern.
+     * Get Ruby specification for arbitrary gem.
+     * @param gempath Full path to gem file or null
+     * @return RubyObject specification
      */
-    public static final Pattern PATH_PATTERN = Pattern.compile("/api/v1/gems/([\\w]+).(json|yml)");
+    JsonObject getinfo(Path gempath);
```
g4s8, line 42: No need to use `get` prefix, `info` is informative enough

src/main/java/com/artipie/gem/GemInfo.java
g4s8: this issue was not fixed

src/main/java/com/artipie/gem/GetGemInfo.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * @since 0.2
+ */
+@SuppressWarnings({"PMD.UnusedPrivateField", "PMD.SingularField"})
+public final class GetGemInfo implements Slice {
+
+    /**
+     * Endpoint path pattern.
+     */
+    public static final Pattern PATH_PATTERN = Pattern.compile("/api/v1/gems/([\\w]+).(json|yml)");
```
g4s8, line 52: Do we support `yml` too?

src/main/java/com/artipie/gem/RubyGemIndex.java
```diff
@@ -47,6 +45,11 @@
      */
     private final Ruby ruby;
 
+    /**
+     * Ruby initialization flag.
+     */
+    private boolean issetup;
```
g4s8, line 51: This field is not thread safe

src/main/java/com/artipie/gem/RubyObjJson.java
```diff
@@ -58,14 +63,15 @@
...
+        this.issetup = false;
     }
 
     /**
      * Create JSON info for gem.
      * @param gempath Full path to gem file or null
      * @return JsonObjectBuilder result
      */
-    public JsonObject createJson(final Path gempath) {
+    public JsonObject getinfo(final Path gempath) {
```
g4s8, line 74: No need to use get prefixes

src/test/java/com/artipie/gem/RubyObjJsonTest.java
```diff
@@ -47,7 +47,8 @@
 public class RubyObjJsonTest {
 
     @Test
-    public void createJsonByPath(@TempDir final Path tmp) throws IOException {
+    public void createJsonByPath(@TempDir final Path tmp) throws IOException,
```
g4s8, line 50: In unit test you can just declare `throws Exception` to avoid listing all possible types here

src/main/java/com/artipie/gem/Gem.java
g4s8: Please try to think yourself how to reduce the coupling between classes

g4s8 (changes requested):
@milkysoft please check my answers

src/main/java/com/artipie/gem/GetGemInfo.java
milkysoft: @g4s8 , as I see calling to routes like `https://rubygems.org/api/v1/gems/gviz.yml` produces no results, only `https://rubygems.org/api/v1/gems/gviz.json` at the end returns data back. Thus for now no urgency to implement yml output, although it is OK to create new Issue for this format

g4s8: @milkysoft yes, this is right direction. In that case, we need to return standard [Not Implemented](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) response for `yml` requests, and fix it later.

src/main/java/com/artipie/gem/RubyObjJson.java
milkysoft: @g4s8 , Fixed in latest commit

src/main/java/com/artipie/gem/RubyGemIndex.java
milkysoft: @g4s8 , Reading and Setting this field fixed in latest commit

src/main/java/com/artipie/gem/GetGemInfo.java
milkysoft: @g4s8 , Fixed in latest commit, corresponding unit test added

src/test/java/com/artipie/gem/RubyObjJsonTest.java
milkysoft: @g4s8 , fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -117,6 +141,57 @@
...
+                        return obj;
+                    }
+                )
+        );
+    }
+
+    /**
+     * Initialize ruby.
+     */
+    public void initialize() {
```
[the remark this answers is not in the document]
milkysoft: @g4s8 , we either initialise in constructor or in separate function. If initialisation is too heavy for constructor, then we consider separate function

src/main/java/com/artipie/gem/GetGemInfo.java
g4s8: @milkysoft I don't see these changes

src/main/java/com/artipie/gem/Gem.java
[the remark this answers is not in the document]
g4s8: Why can't we just lazy initialize ruby interpreter and runtime?

g4s8 (changes requested):
@milkysoft please see new comments and please check a few old not resolved comments too: https://github.com/artipie/gem-adapter/pull/96/files#conversations-menu

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -117,6 +140,60 @@
...
+            tmpdir -> Gem.copyStorage(this.storage, new FileStorage(tmpdir), gem)
+                .thenApply(ignore -> tmpdir)
+        ).thenCompose(
+            tmpdir -> this.sharedInfo()
+                .thenApply(
+                    rubyjson -> {
+                        JsonObject obj = Json.createObjectBuilder().build();
+                        try {
+                            final List<Key> thekey = this.getGemFile(gem).toCompletableFuture()
+                                .get();
```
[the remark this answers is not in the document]
milkysoft: @g4s8 , thenCompose is not applicable here, as inside transfer needed access to many variables, including: `tmpdir`, produced on previous step,  `rubyJson` from `sharedInfo`. If we use thenCompose it produces CompletionStage<CompletionStage> ..., which is incorrect.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -187,4 +266,37 @@ private static Void removeTempDir(final Path tmpdir, final Throwable err) {
...
+            )
+        );
+    }
+
+    /**
+     * Find gem in a given path.
+     * @param gem Gem name to get info
+     * @return String full path to gem file
+     */
+    private CompletionStage<List<Key>> getGemFile(final Key gem) {
```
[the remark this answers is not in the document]
milkysoft: @g4s8 , `gem` here may have the following values: `gviz`, `thor`, etc... Documentation is correct, what has been changed is that instead of getting one value for path, which stucks when gem is absent, the method returns so called "List", actually only one element will be used from this List.
So, filter(...).get(1) has been replaced with List... get(1)

src/main/java/com/artipie/gem/Gem.java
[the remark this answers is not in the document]
milkysoft: @g4s8 , Fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , InterruptedException is produced when User manually interrupts program execution, including typing Ctrl+C or equivalent. These are special cases and if program interrupted manually, all resources could be released manually as well. The code in place does the best job possible to clean up system from temporarily created files and folders.

Issue #104, opened 2021-07-21
Investigate gem naming


g4s8:
Let's check what names uses gem stash to store files on `gem push` and use same naming, now we generate random names on push.

milkysoft:
@g4s8, `gem push` command indeed sends POST request to Ruby Server, which saves attached file into folder similar to ` ~/.gemstash/private/gems/y/a/r/yard_0_9_26-0449f68b65f6eda69d7c1b04abe79594/` for `yard-0.9.26.gem`. In this folder gem file is created together with properties and spec:

```
ls -l ~/.gemstash/private/gems/y/a/r/yard_0_9_26-0449f68b65f6eda69d7c1b04abe79594/
total 752
-rw-rw-r-- 1 zuoqin zuoqin 758272 Jul 21 11:03 gem
-rw-rw-r-- 1 zuoqin zuoqin     49 Jul 21 11:03 properties.yaml
-rw-rw-r-- 1 zuoqin zuoqin    448 Jul 21 11:03 spec
```

g4s8:
> @g4s8, `gem push` command indeed sends POST request to Ruby Server, which saves attached file into folder similar to ` ~/.gemstash/private/gems/y/a/r/yard_0_9_26-0449f68b65f6eda69d7c1b04abe79594/` for `yard-0.9.26.gem`. In this folder gem file is created together with properties and spec:
> 
> ```
> ls -l ~/.gemstash/private/gems/y/a/r/yard_0_9_26-0449f68b65f6eda69d7c1b04abe79594/
> total 752
> -rw-rw-r-- 1 zuoqin zuoqin 758272 Jul 21 11:03 gem
> -rw-rw-r-- 1 zuoqin zuoqin     49 Jul 21 11:03 properties.yaml
> -rw-rw-r-- 1 zuoqin zuoqin    448 Jul 21 11:03 spec
> ```

@milkysoft what is the gem stash root dir in this example?

PR #105, opened 2021-07-21
Issue #104 Rename gem to its name and version

milkysoft:
When a new gem is being pushed with `gem push`, save it not with random name, but use gem name and version

### DIFF
--- a/src/main/java/com/artipie/gem/Gem.java
+++ b/src/main/java/com/artipie/gem/Gem.java
@@ -99,7 +99,12 @@ public final class Gem {
                 .thenApply(ignore -> tmp)
         ).thenCompose(
             tmp -> this.shared.apply(RubyGemIndex::new)
-                .thenAccept(index -> index.update(tmp))
+                .thenAccept(
+                    index -> {
+                        final Path dir = Paths.get(tmp.toString(), gem.string());
+                        index.rename(dir);
+                        index.update(tmp);
+                    })
                 .thenCompose(none -> new Copy(new FileStorage(tmp)).copy(this.storage))
                 .handle(removeTempDir(tmp))
         );
--- a/src/main/java/com/artipie/gem/Gem.java
+++ b/src/main/java/com/artipie/gem/Gem.java
@@ -102,8 +102,7 @@ public final class Gem {
                 .thenAccept(
                     index -> {
                         final Path dir = Paths.get(tmp.toString(), gem.string());
-                        index.rename(dir);
-                        index.update(tmp);
+                        index.update(dir);
                     })
                 .thenCompose(none -> new Copy(new FileStorage(tmp)).copy(this.storage))
                 .handle(removeTempDir(tmp))
--- a/src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
+++ b/src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
@@ -62,4 +62,27 @@ public final class RubyGemIndex implements GemIndex {
             )
         );
     }
+
+    /**
+     * Renames local file.
+     *
+     * @param path Full path too gem file.
+     */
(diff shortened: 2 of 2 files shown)


src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
```diff
@@ -62,4 +62,27 @@ public void update(final Path path) {
...
+        final RubyRuntimeAdapter adapter = JavaEmbedUtils.newRuntimeAdapter();
+        adapter.eval(this.ruby, "require 'rubygems/package.rb'");
+        adapter.eval(
+            this.ruby,
+            String.format(
+                "ind='%s'.rindex('/')\n"
+                    .concat("dir='' if ind == nil \n")
+                    .concat("dir='%s'[0,ind+1] if ind != nil \n")
+                    .concat("spec=Gem::Package.new('%s').spec \n")
+                    .concat("File.rename('%s', dir + spec.name + '-'")
```
g4s8, line 81: What is the point of renaming files in ruby instead of Java?

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -99,7 +99,12 @@ public Gem(final Storage storage) {
                 .thenApply(ignore -> tmp)
         ).thenCompose(
             tmp -> this.shared.apply(RubyGemIndex::new)
-                .thenAccept(index -> index.update(tmp))
+                .thenAccept(
+                    index -> {
+                        final Path dir = Paths.get(tmp.toString(), gem.string());
+                        index.rename(dir);
+                        index.update(tmp);
```
g4s8, line 106: This rename doesn't have effect because it doesn't rename file in the storage, only in local temporary directory

g4s8 (changes requested):

kirill-h5i:
@g4s8 ping

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
```diff
@@ -62,4 +62,27 @@ public void update(final Path path) {
             )
         );
     }
+
+    /**
+     * Renames local file.
+     *
+     * @param path Full path too gem file.
+     */
+    public void rename(final Path path) {
```
g4s8, line 71: @milkysoft I dont think that `rename` is a good method for `Index` class: why index can rename? It's just a metadata parser implementation. So let's keep only related code in this index implementation, e.g. the method to extract spec name from gem

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
```diff
@@ -62,4 +62,27 @@ public void update(final Path path) {
...
+        final RubyRuntimeAdapter adapter = JavaEmbedUtils.newRuntimeAdapter();
+        adapter.eval(this.ruby, "require 'rubygems/package.rb'");
+        adapter.eval(
+            this.ruby,
+            String.format(
+                "ind='%s'.rindex('/')\n"
```
g4s8, line 77: What is `ind`? These ruby lines are too cryptic and hard to understand.

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
g4s8: Remember that this code can be read and maintained by random people from open source who knows java, not ruby. Also, as described above, the rename operation is not logically good for index parser

src/main/java/com/artipie/gem/Gem.java
g4s8: This comment is not valid, since we copy all blobs from tmp storage.

g4s8 (changes requested):
@milkysoft please check my review

PR #106, opened 2021-07-25
Add dependencies route

### LINKED ISSUE #97: Gem dependencies API
Let's investigate what is `/api/v1/dependencies` API and create an SDK method to cover this API request.

milkysoft:
1. Implement /dependencies route
2. Implement /versions route
3. Implement /gems route
4. Make `bundle install` command to succeed

close #97


g4s8:
@milkysoft also, please pay attention to PR style: https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
         this.shared = new SharedRuntime();
     }
 
+    /**
+     * Get info Ruby gem.
```
g4s8, line 100: Why do we need to load Ruby gem into memory?

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+     *
+     * @param filename Ruby file to return
+     * @return Completable action
+     */
+    public CompletionStage<byte[]> getRubyFile(final Key filename) {
```
g4s8, line 105: Key is not a file

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+        )
+        .thenApply(
+            tmpdir -> {
+                final byte[] filecontent;
+                try {
+                    final CompletionStage<Key> stage = this.getGemFile(
+                        filename, true, new Key.From("thor-1.0.1.gemspec.rz")
+                    );
+                    final CompletableFuture<Key> feature = stage.toCompletableFuture();
+                    final Key thekey = feature.get(10, TimeUnit.MILLISECONDS);
```
g4s8, line 123: Please don't use blocking calls, vertx server may stop the request processor on such behavior

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+                    final Path path = Paths.get(tmpdir.toString(), thekey.string());
+                    final File file = new File(path.toString());
+                    filecontent = Files.readAllBytes(file.toPath());
```
g4s8, line 126: Why do you convert path to file and then back to path? You can just use `Files.readAllBytes(path)`

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+                } catch (final IOException | TimeoutException | InterruptedException
```
g4s8, line 127: `InterruptedException` should not be handled in a such way. Please read the documentation about correct `InterruptedException` handling

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+                    | ExecutionException exc) {
+                    throw new ArtipieException(exc);
+                }
+                return filecontent;
+            }
+        ).handle(removeTempDir(dir.get()));
+    }
+
+    /**
+     * Get info Ruby gem.
```
g4s8, line 137: Two same methods?

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+     *
+     * @param filename Ruby file to return
+     * @return Completable action
+     */
+    public CompletionStage<byte[]> getRubyFile(final Key filename) {
+        final AtomicReference<Path> dir = new AtomicReference<>();
+        return newTempDir().thenCompose(
+            tmp -> {
+                dir.set(tmp);
+                return new Copy(this.storage, key -> key.string().contains(filename.string()))
```
g4s8, line 110: `contains` is not correct here because gem name could be in the repository name, e.g. user `rails` may keep package `gviz` in his repository, then the key will be `rails/gviz`

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -89,6 +96,71 @@ public Gem(final Storage storage) {
...
+     *
+     * @param gems Ruby gem to extract info
+     * @return Completable action
+     */
+    public CompletionStage<byte[]> getDependencies(final List<Key> gems) {
+        final AtomicReference<Path> dir = new AtomicReference<>();
+        return newTempDir().thenCompose(
+            tmp -> {
+                dir.set(tmp);
+                return new Copy(this.storage, key -> key.string().contains(gems.get(0).string()))
```
g4s8, line 147: Why we copy only first gem here? And same problem for `contains`

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -157,6 +229,42 @@ public Gem(final Storage storage) {
...
     }
 
+    /**
+     * Find gem in a given path.
+     * @param gem Gem name to get info
+     * @param exact Is exact
+     * @param fallout Is fallout
+     * @return String full path to gem file
+     */
+    private CompletionStage<Key> getGemFile(final Key gem, final boolean exact,
```
g4s8, line 239: The javadoc and method name are confusing here: `getGemFile` returns `Key`, not file. "Find gem in a given path." accepts `Key` not path.

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -157,6 +229,42 @@ public Gem(final Storage storage) {
         );
     }
 
+    /**
+     * Find gem in a given path.
+     * @param gem Gem name to get info
+     * @param exact Is exact
+     * @param fallout Is fallout
```
g4s8, line 236: What is fallout? Method behavior should be clear from Javadoc description

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -157,6 +229,42 @@ public Gem(final Storage storage) {
...
+    private CompletionStage<Key> getGemFile(final Key gem, final boolean exact,
+        final Key fallout) {
+        final String gemstr = ".gem";
+        final CompletableFuture<Key> future = new CompletableFuture<>();
+        Single.fromFuture(this.storage.list(Key.ROOT))
+            .map(
+                list -> {
+                    final List<Key> result;
+                    final Stream<Key> res = list.stream().filter(
+                        key -> key.string().contains(gem.string()) && key.string().endsWith(gemstr)
```
g4s8, line 248: Same here: we can match `rails/gviz.gem` where gem is `rails`

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -157,6 +229,42 @@ public Gem(final Storage storage) {
...
+                            key -> key.string().contains(gem.string())
+                                && key.string().endsWith(gemstr)
+                        ).limit(1).collect(Collectors.toList());
+                    } else {
+                        result = list.stream().filter(
+                            key -> !exact && key.string().contains(gem.string())
+                                || fallout != null && key.string().equals(fallout.string())
+                        ).limit(1).collect(Collectors.toList());
+                    }
+                    return result;
```
g4s8, line 261: This code has too many duplications and perform the same work multiple times - no need to build a stream just to count and then build another stream for filter. 
Also, it's not clear why do you use `List result` which is always containing only one element.

src/main/java/com/artipie/gem/http/ApiGetSlice.java
```diff
@@ -75,18 +91,81 @@ public ApiGetSlice(final Storage storage) {
     public Response response(final String line,
         final Iterable<Map.Entry<String, String>> headers,
         final Publisher<ByteBuffer> body) {
+        final String deproute = "/api/v1/dependencies";
```
g4s8, line 94: No need to put all HTTP handlers into one `Slice`, you can introduce more slices to handle new endpoints. One slice per endpoint

src/main/java/com/artipie/gem/http/ApiGetSlice.java
```diff
@@ -75,18 +91,81 @@ public ApiGetSlice(final Storage storage) {
...
+                    .thenApply(out -> new RsWithBody(ByteBuffer.wrap(out)))
+            );
+        } else {
             throw new IllegalStateException("Invalid routing schema");
         }
+        return res;
+    }
+
+    /**
+     * Process dependency route.
```
g4s8, line 144: Please use meaningful method names and javadocs. Try to imagine that you see this code for the first time, how would you understand what does this method do?

src/main/java/com/artipie/gem/http/ApiGetSlice.java
```diff
@@ -75,18 +91,81 @@ public ApiGetSlice(final Storage storage) {
...
+        final int indexs = line.indexOf(deproute) + offset;
+        final int indexe = line.indexOf(ApiGetSlice.HTTP) - 1;
+        final List<Key> gemkeys = new ArrayList<>(0);
+        for (final String gemname : line.substring(indexs, indexe).split(",")) {
+            gemkeys.add(new Key.From(gemname));
+        }
+        final byte[] obj;
+        try {
+            obj = this.sdk.getDependencies(gemkeys)
+                .toCompletableFuture().get();
```
g4s8, line 161: Don't ever use blocking calls. It's only acceptable in unit tests or some corner cases like futures joining

src/main/java/com/artipie/gem/http/ApiGetSlice.java
```diff
@@ -75,18 +91,81 @@ public ApiGetSlice(final Storage storage) {
...
+        } catch (final InterruptedException | ExecutionException exc) {
```
g4s8, line 162: `InterruptedException` should not be handled in a such way, please check documentation. You can find a very good summary from documentation here: https://stackoverflow.com/a/3976377/1723695

src/main/java/com/artipie/gem/http/GemSlice.java
```diff
@@ -95,6 +95,48 @@ public GemSlice(final Storage storage,
...
+                    ),
+                    new ApiGetSlice(storage)
+                ),
+                new RtRulePath(
+                    new RtRule.All(
+                        new ByMethodsRule(RqMethod.GET),
+                        new RtRule.ByPath("/gems/.*")
+                    ),
+                    new ApiGetSlice(storage)
+                ),
```
g4s8, line 139: Same problem here: no need to handle all endpoints in one slice

src/main/java/com/artipie/gem/ruby/IDependencies.java
```diff
@@ -0,0 +1,38 @@
...
+ * SOFTWARE.
+ */
+package com.artipie.gem.ruby;
+
+/**
+ * Ruby Dependencies gem index implementation.
+ *
+ * @since 1.0
+ */
+public interface IDependencies {
```
g4s8, line 31: 1) why do we need this interface? 2) what does this `I` prefix mean?

src/main/ruby/dependencies.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+require 'rubygems'
+require 'rubygems/package'
+require 'time'
+require 'tmpdir'
+require 'java'
+
+class Dependencies
+    include com.artipie.gem.ruby.IDependencies
```
g4s8, line 9: No need to implement some interfaces here, we don't expose it outside the package. You can just call class method directly as we did previously

src/main/ruby/dependencies.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def initialize(val)
+        @val = val
+    end
+
+    def dependencies()
+        resgems = []
+        @val.split.each do |gem|
+            puts(gem)
```
g4s8, line 18: Please remove stdout logging from production code, you can use only `Logger` from slf4j

g4s8 (changes requested):
@milkysoft please check some initial comments

src/main/ruby/dependencies.rb
milkysoft: @g4s8 , agreed, fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , agreed, fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , agreed, fixed in latest commit

src/main/java/com/artipie/gem/http/GemSlice.java
milkysoft: @g4s8 , agreed, fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , we use the same functionality for all responses, althoough we can reconsider our approach

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , I confirm **Key** here is arbitrary name

src/main/ruby/dependencies.rb
milkysoft: @g4s8 , unfortunately all efforts to convert RubyObject to RubyString failed until now

milkysoft: @g4s8 , fixed in latest commit

src/main/java/com/artipie/gem/ruby/IDependencies.java
milkysoft: @g4s8 , fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
g4s8: @milkysoft but `key` is just a locator of some data in key-value storage, it could be a file or could not

Issue #103, opened 2021-07-19
Enhance output JSON for /api/v1/gems/gemname.json route


milkysoft:
Currently output for `curl  http://host.testcontainers.internal:49199/api/v1/gems/gviz.json` request is:
```
{
   "require_paths":"[\"lib\"]",
   "date":"2015-04-03 00:00:00 UTC",
   "metadata":"{}",
   "signing_key":"",
   "executables":"[\"gviz\"]",
   "rubyforge_project":"",
   "description":"Ruby's interface of graphviz",
   "original_platform":"ruby",
   "platform":"ruby",
   "bindir":"bin",
   "cert_chain":"[]",
   "email":"[\"postagie@gmail.com\"]",
   "summary":"Ruby's interface of graphviz. It generate a dot file with simple ruby's syntax.",
   "requirements":"[]",
   "test_files":"[\"spec/fixtures/graph.ru\", \"spec/gviz_command_spec.rb\", \"spec/gviz_draw_spec.rb\", \"spec/gviz_edge_spec.rb\", \"spec/gviz_node_spec.rb\", \"spec/gviz_spec.rb\", \"spec/spec_helper.rb\", \"spec/system_extension_spec.rb\"]",
   "new_platform":"ruby",
   "required_rubygems_version":">= 0",
   "cache_file":"",
   "version":"0.3.5",
   "dependencies":"[<Gem::Dependency type=:development name=\"rspec\" requirements=\"~> 2.1\">, <Gem::Dependency type=:development name=\"bundler\" requirements=\"~> 1.5\">, <Gem::Dependency type=:development name=\"rake\" requirements=\">= 0\">, <Gem::Dependency type=:runtime name=\"thor\" requirements=\">= 0\">]",
   "extensions":"[]",
   "licenses":"[]",
   "full_name":"",
   "rdoc_options":"[]",
   "specification_version":"4",
   "post_install_message":"",
   "name":"gviz",
[...]

PR #107, opened 2021-07-28
Issue #103 Enhance output JSON

milkysoft:


src/main/java/com/artipie/gem/GemMeta.java
```diff
@@ -50,11 +50,9 @@
     /**
      * Extract Gem info.
      * @param gem Path to gem
-     * @param format Info format
-     * @param <T> Format type
      * @return JSON object
      */
-    <T> T info(Path gem, InfoFormat<T> format);
+    JsonObject info(Path gem);
```
g4s8, line 55: @milkysoft it's wrong change: we may have multiple format outputs: json and yaml. This interface is an abstraction for both types and it could be used to fetch some spec metadata from gem. Json object is too concrete type

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -124,11 +129,9 @@ public Gem(final Storage storage) {
     /**
      * Gem info data.
      * @param gem Gem name
-     * @param fmt Info format
-     * @param <T> Format type
      * @return Future
      */
-    public <T> CompletionStage<T> info(final String gem, final GemMeta.InfoFormat<T> fmt) {
+    public CompletionStage<javax.json.JsonObject> info(final String gem) {
```
g4s8, line 134: And same here: the return type is generic `<T>` here, since we may have different output types for same spec

src/main/ruby/info.rb
```diff
@@ -0,0 +1,45 @@
...
+        spec.authors.each do |author|
+            if authors.length > 0
+                authors = authors + ", "
+            end
+            authors = authors + author
+        end
+        deps = spec.dependencies
+        development = []
+        runtime = []
+        deps.each do |item|
```
g4s8, line 35: What is the reason of making all these transformations in Ruby code? It's very hard to read for Java developers. It doesn't depend on any Ruby internals and could be easily moved to Java code

src/main/java/com/artipie/gem/IRubyInfo.java
```diff
@@ -0,0 +1,40 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package com.artipie.gem;
+
+/**
+ * An SDK, which servers gem packages.
```
g4s8, line 27: Did you try to read your comments before committing changes?

src/main/java/com/artipie/gem/IRubyInfo.java
```diff
@@ -0,0 +1,40 @@
...
+ * <p>
+ * Performes gem index update using specified indexer implementation.
+ * </p>
+ * @since 1.0
+ */
+public interface IRubyInfo {
```
g4s8, line 33: What is the purpose of this interface?

g4s8 (changes requested):

src/main/ruby/info.rb
milkysoft: @g4s8 , theoretically we can obtain some necessary properties for required output JSON with Java code, hovewer parts of required JSON values, such as name, version, authors, etc could be obtained only through utilising Ruby script in some way. Thus, natural implementation would be to obtain all required JSON values at one place, Ruby in this case.

g4s8: > hovewer parts of required JSON values, such as name, version, authors, etc could be obtained only through utilising Ruby script in some way.

@milkysoft it's not true - in previous implementation (current base PR HEAD) it was possible to obtain any Ruby object attribute such as name, version, etc using Java code

g4s8:
Closed by #116

PR #108, opened 2021-07-30
fix: corrected gem naming on index

g4s8:
Extracting gem name and version from metadata,
compose actual key of gem using `<name>-<version>.gem`.
Fixed and refactored SDK test.

Fix: #104
Close: #105


olenagerasimova (approved):
@g4s8 thanks, looks good to me

Issue #110, opened 2021-08-02
Copy all gems for indexing instead of only one


milkysoft:
When gem-adapter performs indexing, it requires that all server-side gems would be available for indexing as `specs` along with all other files being recreated

g4s8:
We don't need to copy all gem repository for indexing, since the repository is usually huge - it can't be placed on server instance temporary dir, the download and upload operation are very expensive, it requires transferring terabytes of data via network.

PR #111, opened 2021-08-02
Issue #110 Create meta data for all available gems when indexing

milkysoft:


g4s8 (changes requested):
@milkysoft this approach is not acceptable - see ticket comment: https://github.com/artipie/gem-adapter/issues/110#issuecomment-892448693

PR #112, opened 2021-08-02
Delete temporarily saved gem

### LINKED ISSUE #109: Remove temporarily files under gem directory
When gem adapter accepts `gem push` request, it removes temporarily directory when indexing finishes. In addition, temporarily files should be removed from `gems` folder in current directory as well.

milkysoft:
close #109 

Remove temporarily saved files to create Ruby metadata before sending back HTTP response.

### DIFF
--- a/src/main/java/com/artipie/gem/http/SubmitGemSlice.java
+++ b/src/main/java/com/artipie/gem/http/SubmitGemSlice.java
@@ -72,8 +72,11 @@ public final class SubmitGemSlice implements Slice {
         return new AsyncResponse(
             this.storage.save(
                 key, new ContentWithSize(body, headers)
-            ).thenCompose(none -> this.gem.update(key))
-                .thenApply(none -> new RsWithStatus(RsStatus.CREATED))
+            ).thenCompose(none -> this.gem.update(key)).thenApply(
+                none -> {
+                    this.storage.delete(key);
+                    return new RsWithStatus(RsStatus.CREATED);
+                })
         );
     }
 }
--- a/src/main/java/com/artipie/gem/http/SubmitGemSlice.java
+++ b/src/main/java/com/artipie/gem/http/SubmitGemSlice.java
@@ -73,8 +73,8 @@ public final class SubmitGemSlice implements Slice {
             this.storage.save(
                 key, new ContentWithSize(body, headers)
             ).thenCompose(none -> this.gem.update(key))
-                .thenApply(none -> this.storage.delete(key))
-                .thenApply(none -> new RsWithStatus(RsStatus.CREATED))
+            .thenCompose(none -> this.storage.delete(key))
+            .thenApply(none -> new RsWithStatus(RsStatus.CREATED))
         );
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/com/artipie/gem/http/SubmitGemSlice.java
```diff
@@ -72,8 +72,11 @@ public Response response(final String line, final Iterable<Entry<String, String>
         return new AsyncResponse(
             this.storage.save(
                 key, new ContentWithSize(body, headers)
-            ).thenCompose(none -> this.gem.update(key))
-                .thenApply(none -> new RsWithStatus(RsStatus.CREATED))
+            ).thenCompose(none -> this.gem.update(key)).thenApply(
+                none -> {
+                    this.storage.delete(key);
```
g4s8, line 77: `this.storage.delete(key);` is async call, we need to chain async calls, not just forget about it. Also, I'd change the key of this blob from `gems/<uuid>` to `_uploads/<uuid>` to indicate that it's a temporary upload prefix and not mix it with `gems`

g4s8 (changes requested):
@milkysoft thanks, please check my comment

g4s8:
@milkysoft another comment about PR style: please don't forget to provide the description, and add links to the PR body, not title - title issue link does nothing, but description link creates a reference in GitHub between issue and PR. See https://github.com/artipie/artipie/blob/master/CONTRIBUTING.md

src/main/java/com/artipie/gem/http/SubmitGemSlice.java
milkysoft: @g4s8 , chaining implemented in latest commit.

Regarding changing directory, most probably indexing will not work if subdirectory has name other then `gems`. An idea is that Ruby indexing creates Ruby metadata in the current folder and Marshal subfolders with content for all gems stored in gems subfolder, not any other.

src/main/java/com/artipie/gem/http/SubmitGemSlice.java
```diff
@@ -73,6 +73,7 @@ public Response response(final String line, final Iterable<Entry<String, String>
             this.storage.save(
                 key, new ContentWithSize(body, headers)
             ).thenCompose(none -> this.gem.update(key))
+                .thenApply(none -> this.storage.delete(key))
```
g4s8, line 76: It's still not fixed: `storage.delete(key)` returns future, method `thenApply` maps the result of previous call to lambda result. So you're just ignoring the result of `delete()` future. To fix it, use `thenCompose` instead of `thenApply`

g4s8 (changes requested):
@milkysoft the concurrency problem was not fixed

src/main/java/com/artipie/gem/http/SubmitGemSlice.java
milkysoft: @g4s8 , Understood & agreed. `thenApply` is replaced withn `thenCompose` in latest commit

g4s8 (approved):
@milkysoft thanks

PR #116, opened 2021-08-06
Use Java to create Gem info output JSON

milkysoft:


g4s8:
@milkysoft why do we need a new pull request if we have #107 ?
This PR looks like a fixes for some review comments in #107 - so please, continue making changes in #107 and close this one to keep review history and to avoid noise in repository.

src/main/java/com/artipie/gem/GemMeta.java
```diff
@@ -69,6 +60,6 @@
          * @param data Gem info data
          * @return Formatted info
          */
-        T print(Map<String, String> data);
+        T print(JsonObject data);
```
g4s8, line 63: It's abstraction violation - this class is an intermediate layer for converting abstract gem data into concrete data format. Don't use json as parameter of this method. Please check review comments to previous PR for info response improvements, it was one of the major comment there

src/main/java/com/artipie/gem/ruby/RubyGemMeta.java
```diff
@@ -54,20 +77,176 @@ public RubyGemMeta(final Ruby ruby) {
...
+                builder.add("authors", jsonauthors);
+            } else if (hstr.equals(thevar.getName())) {
+                builder.add("homepage", RubyGemMeta.getVar(vars, hstr));
+            } else if (vstr.equals(thevar.getName())) {
+                builder.add("version", RubyGemMeta.getVar(vars, vstr));
+            } else if ("@dependencies".equals(thevar.getName())) {
+                final JsonObjectBuilder jsondep = Json.createObjectBuilder();
+                this.getDependencies(adapter, gem, jsondep);
+                builder.add("dependencies", jsondep);
+            }
```
g4s8, line 106: Why we can't just take **all** variables from spec metadata and render them into required format? Current implementation covers much more variables from spec metadata. In case if you need to filter out some attributes, you can just use some set as a filter for white-list or black-list attributes filtering.

src/main/java/com/artipie/gem/ruby/RubyGemMeta.java
```diff
@@ -54,20 +77,176 @@ public RubyGemMeta(final Ruby ruby) {
 
     @Override
     public <T> T info(final Path gem, final GemMeta.InfoFormat<T> fmt) {
+        final String hstr = "@homepage";
+        final String vstr = "@version";
         final RubyRuntimeAdapter adapter = JavaEmbedUtils.newRuntimeAdapter();
+        final JsonObjectBuilder builder = Json.createObjectBuilder();
```
g4s8, line 83: See previous comment: do not construct JSON on this abstraction layer

src/main/java/com/artipie/gem/ruby/RubyGemMeta.java
```diff
@@ -54,20 +77,176 @@ public RubyGemMeta(final Ruby ruby) {
...
+     *
+     * @param varos For variables
+     * @param name For variable name
+     * @return String variable value
+     */
+    private static String getVar(final List<Variable<Object>> varos, final String name) {
+        String res = "";
+        for (final Variable<Object> var : varos) {
+            if (name.equals(var.getName())) {
+                res = var.getValue().toString();
```
g4s8, line 123: Why always a string? JSON has more types, rubygems uses numbers as well as strings, see https://guides.rubygems.org/rubygems-org-api/

g4s8 (changes requested):
@milkysoft please check a few comments. Also, don't forget to fix PR description and check review comments for previous PR, since some of them are the same

src/main/java/com/artipie/gem/ruby/RubyGemMeta.java
milkysoft: @g4s8 , sure we will support all required types. This code uses `toString` method to return value from RubyVariable, it is possible to do only in String format. As alternative, we can consider independent Ruby interfaces which will read values from corresponding Ruby scripts

src/main/java/com/artipie/gem/GemMeta.java
milkysoft: @g4s8 , I agree to add abstraction. The only change is that abstraction returns now Json tree instead of previously used Map

src/main/java/com/artipie/gem/ruby/RubyGemMeta.java
milkysoft: @g4s8 , this question relates to the following contradiction, so neither answer would be correct or incorrect. The task is to build output JSON which closely matches format from rubygems.org as in https://api.rubygems.org/api/v1/gems/gviz.json. This format contains values, which could be obtained through `Gem::Package.new('%s').spec`, but also a lot of other values, such as `sha` checksum, which should be obtained through other techniques. Also each value from `spec` should relates to specific key, which has custom logic, for example, while is `spec` there is `homepage` variable, output JSON should have `homepage_uri` JSON key at root.

g4s8: @milkysoft I see, but why do we need to check each attribute name with this if-else block if we can but all of them to some set and check if set contains this variable or not?

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -103,7 +104,21 @@ public Gem(final Storage storage) {
                 .thenApply(
                     meta -> meta.info(
                         Paths.get(tmp.toString(), gem.string()),
-                        spec -> String.format("%s-%s.gem", spec.get("name"), spec.get("version"))
+                        spec -> {
+                            String name = "";
+                            String version = "";
+                            for (final TreeNode<ImmutablePair<String, String>> node : spec) {
```
olenagerasimova, line 110: `ImmutablePair` is `Pair` implementation, let's use higher abstraction (`Pair`) for variables, method return types etc and `ImmutablePair` when object creation is performed, e.g.:
```
for (final TreeNode<Pair<String, String>> node : spec) { ... }

Collection<Pair<String, String>> pairs = new ListOf<>(new ImmutablePair<>("left", "right"))
```

src/main/java/com/artipie/gem/TreeNodeIter.java
```diff
@@ -0,0 +1,136 @@
...
+package com.artipie.gem;
+
+import java.util.Iterator;
+
+/**
+ * Gem metadata parser.
+ * @param <T> Format type
+ * @since 1.0
+ */
+@SuppressWarnings("PMD.NullAssignment")
```
olenagerasimova, line 34: this suppression seems unnecessary

olenagerasimova (changes requested):
@milkysoft please, check my comments. Also, my main and general concern here is the lack of unit tests. All classes (`TreeNode`, `TreeNodeIter` and others) should be unit-tested. Not in this PR (it's already to big), but in the next one, when this one will be merged.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @olenagerasimova , the problem is the following: Pair is an abstract, cannot be instantiated, so you propose to instantiate ImmutablePairs, correct. But we use TreeNode<ImmutablePair>, not List.

if we replace ImmutablePair with Pair in:
`T print(TreeNode<ImmutablePair<String, String>> data);`


then in RubyGemMeta, where we instantiate ImuutablePair https://github.com/milkysoft/gem-adapter/blob/796f22e5824233862bfcf3cf21fd8ff2cf8d2e96/src/main/java/com/artipie/gem/ruby/RubyGemMeta.java#L141:

`return fmt.print(root);`

wil not compile. Thus declarations should always correspond to implementations

src/main/java/com/artipie/gem/TreeNodeIter.java
milkysoft: @olenagerasimova , agreed, fixed in latest commit

olenagerasimova:
@milkysoft thanks
@g4s8 all my concerns are fixed here, please, take a look/merge

g4s8 (changes requested):
The code in this [blog post](https://berativ.com/data-structures-2/) is not published under any public license, we can't just take and use it in our project because it could lead to serious copyright problems.

milkysoft:
@g4s8 , as an alternative to the blog post link, we can consider taking the code from https://github.com/ctapmex/zkTreeUtil/tree/master/src/main/java/com/dobrunov/zktreeutil publicky available repository, right?

g4s8:
@milkysoft please check #123 - I added new format interfaces for tree structure to support JSON and YAML formats there.

PR #119, opened 2021-08-11
Fix Bug: Gem index is replaced on update bug

### LINKED ISSUE #114: Bug: Gem index is replaced on update
When a user is updating Gem repository metadata using SDK, the SDK replaces old metadata with new one for single package uploaded. The metadata files should be updated, and new gem information should be appended to old files instead of replacing.

We need to find a method to update metadata **without copying** the full repository from storage to local machine.

milkysoft:
close #114 

When a new gem is being pushed, `latest_spec` and `spec` add information for a new gem, keep existing data without rewrite.

### DIFF
--- a/src/main/java/com/artipie/gem/Gem.java
+++ b/src/main/java/com/artipie/gem/Gem.java
@@ -28,7 +28,6 @@ import com.artipie.asto.Copy;
 import com.artipie.asto.Key;
 import com.artipie.asto.Storage;
 import com.artipie.asto.fs.FileStorage;
-import com.artipie.asto.misc.UncheckedConsumer;
 import com.artipie.gem.ruby.RubyGemIndex;
 import com.artipie.gem.ruby.RubyGemMeta;
 import com.artipie.gem.ruby.SharedRuntime;
@@ -112,7 +111,7 @@ public final class Gem {
                         try {
                             Files.move(path, target);
                         } catch (final IOException err) {
-                            throw new ArtipieException(err);
+                            throw new ArtipieIOException(err);
                         }
                         return target;
                     }
--- a/src/main/java/com/artipie/gem/Gem.java
+++ b/src/main/java/com/artipie/gem/Gem.java
@@ -105,19 +105,27 @@ public final class Gem {
                         Paths.get(tmp.toString(), gem.string()),
                         spec -> String.format("%s-%s.gem", spec.get("name"), spec.get("version"))
                     )
-                ).thenAccept(
-                    new UncheckedConsumer<>(
-                        name -> {
-                            final Path path = Paths.get(tmp.toString(), gem.string());
-                            Files.move(path, path.getParent().resolve(name));
+                ).thenApply(
+                    name -> {
+                        final Path path = Paths.get(tmp.toString(), gem.string());
+                        final Path target = path.getParent().resolve(name);
+                        try {
+                            Files.move(path, target);
+                        } catch (final IOException err) {
+                            throw new ArtipieException(err);
                         }
-                    )
-                ).thenApply(none -> tmp)
+                        return target;
+                    }
+                )
         ).thenCompose(
             tmp -> this.shared.apply(RubyGemIndex::new)
                 .thenAccept(index -> index.update(tmp))
-                .thenCompose(none -> new Copy(new FileStorage(tmp)).copy(this.storage))
-                .handle(removeTempDir(tmp))
+                .thenCompose(
+                    none -> new Copy(
+                        new FileStorage(tmp.getParent().getParent())
+                    ).copy(this.storage)
+                )
+                .handle(removeTempDir(tmp.getParent().getParent()))
         );
     }
 
--- a/src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
+++ b/src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
@@ -24,8 +24,12 @@
 
 package com.artipie.gem.ruby;
 
+import com.artipie.ArtipieException;
 import com.artipie.gem.GemIndex;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
+import org.apache.commons.io.IOUtils;
 import org.jruby.Ruby;
 import org.jruby.RubyRuntimeAdapter;
 import org.jruby.javasupport.JavaEmbedUtils;
@@ -53,12 +57,28 @@ public final class RubyGemIndex implements GemIndex {
     @Override
     public void update(final Path path) {
         final RubyRuntimeAdapter adapter = JavaEmbedUtils.newRuntimeAdapter();
+        try {
+            final String script = IOUtils.toString(
+                RubyGemIndex.class.getResourceAsStream(String.format("/metarunner.rb")),
+                StandardCharsets.UTF_8
+            );
+            adapter.eval(this.ruby, script);
+            JavaEmbedUtils.invokeMethod(
+                this.ruby,
+                adapter.eval(this.ruby, "MetaRunner"),
+                "new",
+                new Object[]{path.toString()},
+                Object.class
+            );
+        } catch (final IOException err) {
+            throw new ArtipieException(err);
+        }
         adapter.eval(this.ruby, "require 'rubygems/indexer.rb'");
         adapter.eval(
             this.ruby,
             String.format(
                 "Gem::Indexer.new('%s', {build_modern:true}).generate_index",
-                path.toAbsolutePath().toString()
+                path.getParent().getParent().toAbsolutePath()
             )
         );
     }
--- a/src/main/ruby/metarunner.rb
+++ b/src/main/ruby/metarunner.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+require 'updatemeta.rb'
+
+class MetaRunner
+
+    def initialize(val)
+        Ex.new(val).info()
+        puts val
+    end
+end
--- a/src/main/ruby/updatemeta.rb
+++ b/src/main/ruby/updatemeta.rb
@@ -3,35 +3,45 @@ require 'rubygems'
 require 'rubygems/package'
 require 'time'
 require 'tmpdir'
-require 'digest'
-require 'json'
 require 'java'
 
 class Ex
-    include com.artipie.gem.IRubyInfo
-
     def initialize(val)
         @val = val
     end
 
     def info()
-        spec = Gem::Package.new(@val).spec
-        metas = Marshal.load(File.open("latest_specs.4.8").read)
-        found = false
-        metas.each do |item|
-            if item[0] == spec.name && item[1].version == spec.version.version
-                found = true
-                puts 'Found specification'
+        puts "222222"
+        puts @val
+        begin
+            spec = Gem::Package.new(@val).spec()
+            found = false
+            metas = []
+            if File.file?("latest_specs.4.8") and File.size("latest_specs.4.8") > 10
+                puts "Check latest"
+                content = File.open("latest_specs.4.8").read
+                puts content
+                metas = Marshal.load(content)
+                metas.each do |item|
+                    if item[0] == spec.name && item[1].version == spec.version.version
+                        found = true
+                        puts 'Found specification'
+                    end
+                end
             end
-        end
-        if found == false
-            puts 'Did not found specification'
-            metas.push([spec.name, Gem::Version.create(spec.version.version), "ruby"])
-            data = Marshal.dump(metas)
-            File.write('latest_specs.4.8', data)
-            Zlib::GzipWriter.open('latest_specs.4.8.gz') do |gz|
-                gz.write data
+            if found == false
+                puts 'Did not found specification'
+                metas.push([spec.name, Gem::Version.create(spec.version.version), "ruby"])
+                data = Marshal.dump(metas)
+                File.write('latest_specs.4.8', data)
+                Zlib::GzipWriter.open('latest_specs.4.8.gz') do |gz|
+                    gz.write data
+                end
+                puts "Final spec"
+       	        puts data
             end
+        rescue Exception => e
+            puts e
         end
     end
 end
--- a/src/main/ruby/updatemeta.rb
+++ b/src/main/ruby/updatemeta.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+require 'rubygems'
+require 'rubygems/package'
+require 'time'
+require 'tmpdir'
+require 'digest'
+require 'json'
+require 'java'
+
+class Ex
+    include com.artipie.gem.IRubyInfo
+
+    def initialize(val)
+        @val = val
+    end
+
+    def info()
+        sha = Digest::SHA2.new
+        File.open(@val) do |f|
+            while chunk = f.read(256)
+                sha << chunk
+            end
+        end
+        spec = Gem::Package.new(@val).spec
+        authors = ""
+        spec.authors.each do |author|
+            if authors.length > 0
+                authors = authors + ", "
+            end
+            authors = authors + author
+        end
+        deps = spec.dependencies
+        development = []
+        runtime = []
+        deps.each do |item|
+            if item.type == :runtime
+                runtime.push({'name' => item.name, 'requirements' => item.requirements_list()[0]})
+            else
+                development.push({'name' => item.name, 'requirements' => item.requirements_list()[0]})
+            end
+        end
+        info = {'name' => spec.name, 'version' => spec.version.version, 'sha' => sha.hexdigest, 'info' => spec.description, 'authors' => authors, 'homepage_uri' => spec.homepage, 'dependencies' => {'development' => development, 'runtime' => runtime}}
+        return info.to_json
+    end
+end
--- a/src/main/ruby/updatemeta.rb
+++ b/src/main/ruby/updatemeta.rb
@@ -15,31 +15,23 @@ class Ex
     end
 
     def info()
-        sha = Digest::SHA2.new
-        File.open(@val) do |f|
-            while chunk = f.read(256)
-                sha << chunk
-            end
-        end
         spec = Gem::Package.new(@val).spec
-        authors = ""
-        spec.authors.each do |author|
-            if authors.length > 0
-                authors = authors + ", "
-            end
-            authors = authors + author
-        end
-        deps = spec.dependencies
-        development = []
-        runtime = []
-        deps.each do |item|
-            if item.type == :runtime
-                runtime.push({'name' => item.name, 'requirements' => item.requirements_list()[0]})
-            else
-                development.push({'name' => item.name, 'requirements' => item.requirements_list()[0]})
+        metas = Marshal.load(File.open("latest_specs.4.8").read)
+        found = false
+        metas.each do |item|
+            if item[0] == spec.name && item[1].version == spec.version.version
+                found = true
+                puts 'Found specification'
             end
         end
-        info = {'name' => spec.name, 'version' => spec.version.version, 'sha' => sha.hexdigest, 'info' => spec.description, 'authors' => authors, 'homepage_uri' => spec.homepage, 'dependencies' => {'development' => development, 'runtime' => runtime}}
-        return info.to_json
+        if found == false
+            puts 'Did not found specification'
+            metas.push([spec.name, Gem::Version.create(spec.version.version), "ruby"])
+            data = Marshal.dump(metas)
+            File.write('latest_specs.4.8', data)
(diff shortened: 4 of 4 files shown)


src/main/java/com/artipie/gem/Gem.java
```diff
@@ -105,19 +104,27 @@ public Gem(final Storage storage) {
...
+                        try {
+                            Files.move(path, target);
+                        } catch (final IOException err) {
+                            throw new ArtipieIOException(err);
                         }
-                    )
-                ).thenApply(none -> tmp)
+                        return target;
+                    }
+                )
```
g4s8, line 118: No need in this change, previous version does exactly the same - throwing ArtipieException incase of exception in `move`

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -105,19 +104,27 @@ public Gem(final Storage storage) {
...
         ).thenCompose(
             tmp -> this.shared.apply(RubyGemIndex::new)
                 .thenAccept(index -> index.update(tmp))
-                .thenCompose(none -> new Copy(new FileStorage(tmp)).copy(this.storage))
-                .handle(removeTempDir(tmp))
+                .thenCompose(
+                    none -> new Copy(
+                        new FileStorage(tmp.getParent().getParent())
```
g4s8, line 124: It's not correct - `tmp.getParent()` refers out of sandbox directory and contain other temp files create by other application, not relevant to Artipie

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
```diff
@@ -53,13 +57,21 @@ public RubyGemIndex(final Ruby ruby) {
...
-        adapter.eval(this.ruby, "require 'rubygems/indexer.rb'");
-        adapter.eval(
-            this.ruby,
-            String.format(
-                "Gem::Indexer.new('%s', {build_modern:true}).generate_index",
-                path.toAbsolutePath()
-            )
-        );
+        try {
+            final String script = IOUtils.toString(
```
g4s8, line 61: No need to use script here, the only line in this script is `Gem::Indexer.new(dir2, {build_modern:true}).generate_index`, all other file operations could be easilly done in Java before evaluating this line`

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+require 'rubygems'
+require 'rubygems/package'
+require 'time'
+require 'java'
+
+class Ex
+    def initialize(val)
+        @val = val
+    end
+
```
g4s8, line 11: No need to create a Ruby class here - a method is enough

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+    def info()
+        begin
+            spec = Gem::Package.new(@val).spec()
+            found = false
+            metas = []
+            f = 'latest_specs.4.8'
+            if File.file?(f) and File.size(f) > 10
```
g4s8, line 18: We can check that file exists before calling this script in Java to avoid big Ruby runtime evaluation overhead, calling each line here is very expensive operation

g4s8: And what is this `10` number? why not `5` or `100`?

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+                content = File.open(f).read
+                metas = Marshal.load(content)
+                metas.each do |item|
+                    if item[0] == spec.name && item[1].version == spec.version.version
+                        found = true
+                        puts 'Found specification in latest'
```
g4s8, line 24: Please remove debug message from final code

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+                    end
+                end
+            end
+            if found == false
```
g4s8, line 28: And what happens if spec was not found? We won't write any file then?

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+                puts 'Did not found specification in latest'
```
g4s8, line 29: Same here - no need to print debug messages

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+            end
+            if found == false
+                puts 'Did not found specification in specs'
+                metas.push([spec.name, Gem::Version.create(spec.version.version), "ruby"])
+                data = Marshal.dump(metas)
+                File.write(f, data)
+                Zlib::GzipWriter.open(f + '.gz') do |gz|
+                    gz.write data
+                end
+            end
```
g4s8, line 59: This code is exactly the same as previous, it culd be refactored

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,64 @@
...
+        rescue Exception => e
+            puts e
+        end
```
g4s8, line 62: No need to hide the exception here, let's remove this catch block to allow it blows up to the top level handlers.

g4s8 (changes requested):
@milkysoft please check my comment

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , there are two case I carefully thought about:

1. Client pushed gem, which has been previously pushed and meta data generated;
2. Client pushed a new gem, which has never been pushed before and no meta data exists.


In first case, meta data does not changed at all, we return existing meta data, including that for gem being pushed back.
In second case, a new meta data is generated, including newly pushed gem, then this newly generated meta data is returned back to the client

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , there are two possibilities to use `Files.move(path, target);`:

1. Add try/catch block;
2. Add exception to method signature.

Previously used `thenAccept` has been replaced with `thenApply`, which caused trouble to process exception.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , this code is so great that it is hard to estimate from the first glance. Previously `tmp` variable referenced the directory, where the new meta data should be created using Gem::Indexer Ruby package.

Now, `tmp` variable refers to the full path for the gem, for which newly used meta data should be extracted and then appended to the newly built meta data before returning back to client.

In order words, suppose we are working with `/tmp/Gem9887676576` directory, then now `tmp` variable will be ``/tmp/Gem9887676576/gems/newly-added-1.0.1.gem`, and `tmp.getParent().getParent()` code will return correctly `/tmp/Gem9887676576` path

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , we have approved pattern for JRuby interoperability, which uses class, also `JavaEmbedUtils.invokeMethod(` needed to specify method in the call, which is possible only for Ruby class instead of simple method.

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , 10 is arbitrary selected, 100 is big enough, 5 would be more than OK too.

@g4s8 , it is assumed that our software runs on the best available servers in the industry, thus any discrepancy in Ruby vs Java input-output performance is minor factor.

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , agreed, removed in latest commit

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , agreed, removed in latest commit

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , agreed, removed in latest commit

src/main/java/com/artipie/gem/Gem.java
g4s8: @milkysoft previously it was handled by `UncheckedConsumer`, so you can just revert your changes in this file and everything will be fine

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
milkysoft: @g4s8 , I worry you underestimate complexity to generate Ruby meta data using Java functions. You can check how complexity of output increases when the number of containing elements increases.

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , agreed, fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
g4s8: @milkysoft  I see, in that case we need to remove origin `tmp` and give other usages of `tmp` better names, no need to do magic with parent of parent

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
g4s8: I'm talking about this script:
```
    def initialize(val)
        Ex.new(val).info()
        dir1 = File.dirname(val)
        dir2 = File.expand_path("..", dir1)
        Gem::Indexer.new(dir2, {build_modern:true}).generate_index
        FileUtils.cp('latest_specs.4.8', dir2)
        FileUtils.cp('latest_specs.4.8.gz', dir2)
        FileUtils.cp('specs.4.8', dir2)
        FileUtils.cp('specs.4.8.gz', dir2)
```
The only string here that really should be written on ruby is 
```
        Gem::Indexer.new(dir2, {build_modern:true}).generate_index
```

milkysoft: @g4s8 , understood, agreed, fixed in latest commit.

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , understood, fixed in latest commit

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , `tmp` variable renamed to `fullpath` with latest commit

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -105,19 +105,25 @@ public Gem(final Storage storage) {
...
         ).thenCompose(
-            tmp -> this.shared.apply(RubyGemIndex::new)
-                .thenAccept(index -> index.update(tmp))
-                .thenCompose(none -> new Copy(new FileStorage(tmp)).copy(this.storage))
-                .handle(removeTempDir(tmp))
+            fullpath -> this.shared.apply(RubyGemIndex::new)
+                .thenAccept(index -> index.update(fullpath))
+                .thenCompose(
+                    none -> new Copy(
+                        new FileStorage(fullpath.getParent().getParent())
```
g4s8, line 123: The problem was not fixed - we're still using `getParent().getParent()` which could be hard to understand. We renamed the variable to make it possible to use `tmp` directory

src/main/java/com/artipie/gem/Gem.java
```diff
@@ -105,19 +105,25 @@ public Gem(final Storage storage) {
...
+                    ).copy(this.storage)
+                )
+                .handle(removeTempDir(fullpath.getParent().getParent()))
```
g4s8, line 126: Same here

src/main/ruby/metarunner.rb
```diff
@@ -0,0 +1,14 @@
...
+class MetaRunner
+
+    def initialize(val)
+        Ex.new(val).info()
+        dir1 = File.dirname(val)
+        dir2 = File.expand_path("..", dir1)
+        puts 'Directory: ' + File.expand_path(File.dirname(File.dirname(__FILE__)))
+        Gem::Indexer.new(dir2, {build_modern:true}).generate_index
+    end
+end
```
g4s8, line 14: Why do we need two script files if all the code could be merged into single one?

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
```diff
@@ -53,13 +74,28 @@ public RubyGemIndex(final Ruby ruby) {
...
+            adapter.eval(this.ruby, script);
+            JavaEmbedUtils.invokeMethod(
+                this.ruby,
+                adapter.eval(this.ruby, "MetaRunner"),
+                "new",
+                new Object[]{path.toString()},
+                Object.class
+            );
+            for (final String file : RubyGemIndex.META_NAMES) {
+                final Path source = Paths.get(Paths.get("").toAbsolutePath().toString(), file);
```
g4s8, line 91: `Paths.get("")` returns current working directory, we can't touch this directory, since it could be read-only. Also, please try to make code easier to understand, e.g. this line could be replaced with `path.resolve(file)`

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
```diff
@@ -53,13 +74,28 @@ public RubyGemIndex(final Ruby ruby) {
...
+                final Path target = Paths.get(
+                    path.getParent().getParent().toAbsolutePath().toString(), file
+                );
```
g4s8, line 94: It's very hard to understand what you are doing here

src/main/ruby/metarunner.rb
```diff
@@ -0,0 +1,14 @@
...
+require 'updatemeta.rb'
+require 'rubygems/indexer.rb'
+
+class MetaRunner
+
+    def initialize(val)
+        Ex.new(val).info()
+        dir1 = File.dirname(val)
+        dir2 = File.expand_path("..", dir1)
+        puts 'Directory: ' + File.expand_path(File.dirname(File.dirname(__FILE__)))
```
g4s8, line 11: Please don't print to stdout. And this log message doesn't make any sense for production code

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def info()
+        spec = Gem::Package.new(@val).spec()
+        metas = []
+        metafiles = ['latest_specs.4.8', 'specs.4.8']
+        metafiles.each do |f|
+            metas = []
+            found = false
+            if File.file?(f) and File.size(f) > 10
```
g4s8, line 19: Do we really need to check the file size? What it can say to us? Why it's not enough to check that file is exist

g4s8 (changes requested):

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , `tmp` variable is created at first step, in order to use this variable at the last step when this directory is being deleted, it should be returned from all intermediary transformation steps. But, there is intermediary transformation which needs not only path to temporarily directory, but also full path to submitted gem. Thus, this step needs to receive full path, not only temporarily directory. At this transformation `tmp` is being "lost" and instead, full path is available. We can declare new variable and assign it `fullpath.getparent().getparent()` value as a solution

src/main/java/com/artipie/gem/Gem.java
milkysoft: @g4s8 , `tmp` variable is created at first step, in order to use this variable at the last step when this directory is being deleted, it should be returned from all intermediary transformation steps. But, there is intermediary transformation which needs not only path to temporarily directory, but also full path to submitted gem. Thus, this step needs to receive full path, not only temporarily directory. At this transformation `tmp` is being "lost" and instead, full path is available. We can declare new variable and assign it `fullpath.getparent().getparent()` value as a solution

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , agreed, file size check removed in latest commit

src/main/ruby/metarunner.rb
milkysoft: @g4s8 , understood, fixed in latest commit

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
milkysoft: @g4s8 , an idea of this code is the following:

1. We have persistent storage on server side where all meta data for all submitted gems is stored. Usually, it is current directory;
2. This fragment gets path to each meta data related file in current directory;

src/main/ruby/metarunner.rb
milkysoft: @g4s8 , Ruby scripts combined in one file with latest commit

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
g4s8: @milkysoft current directory could be **READ ONLY**

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
milkysoft: @g4s8 , `.getParent().getParent()` removed in latest commit

src/main/java/com/artipie/gem/ruby/RubyGemIndex.java
milkysoft: @g4s8 , understood, all file system manipulations are performing in temporary directory since last commit

src/main/ruby/metarunner.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+require 'rubygems/indexer.rb'
+
+class MetaRunner
+
+    def initialize(val)
+        dir1 = File.dirname(val)
+        dir2 = File.expand_path("..", dir1)
```
g4s8, line 8: What does `dir1` and `dir2` mean? Let's use self-descriptive variable names.

g4s8: Also, don't rely on current working directory, and it's parent `..`, always assume it's read-only

src/main/ruby/metarunner.rb
```diff
@@ -0,0 +1,47 @@
...
+                metas.push([spec.name, Gem::Version.create(spec.version.version), "ruby"])
+                data = Marshal.dump(metas)
+                metadata.push(data)
+            else
+                metadata.push(content)
+            end
+        end
+
+        Gem::Indexer.new(dir2, {build_modern:true}).generate_index
+        ind = 0
```
g4s8, line 37: this variable looks redundant - it's only updated but not used

src/main/ruby/updatemeta.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
```
g4s8, line 1: This file is not used anymore and could be removed

g4s8 (changes requested):
@milkysoft please check a few minor comments. Also, I don't see any test to reproduce the bug #114 - how we can check that metadata is updated correctly after this fix?

src/main/ruby/metarunner.rb
milkysoft: @g4s8 , please, find `ind` variable usages:

image

src/main/ruby/updatemeta.rb
milkysoft: @g4s8 , understood, agreed

src/main/ruby/metarunner.rb
milkysoft: @g4s8 , `dir1` and `dir2` renamed correspondingly in latest commit

milkysoft:
> @milkysoft please check a few minor comments. Also, I don't see any test to reproduce the bug #114 - how we can check that metadata is updated correctly after this fix?

@g4s8 , corresponding unit test for updated meta data file size added with latest commit

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -112,6 +113,14 @@ public void gemPushAndInstallWorks(@TempDir final Path temp, @TempDir final Path
...
             Matchers.equalTo(0)
         );
+        MatcherAssert.assertThat(
+            String.format("file size is wrong"),
+            this.bash(
+                ruby,
+                String.format("var=$(stat -c%%s latest_specs.4.8); (exit $var)")
+            ),
+            Matchers.equalTo(latestsize)
+        );
```
g4s8, line 123: @milkysoft This code looks really weird - what is the strange way to get file size in Java? But it's not important here, because the file size doesn't say about metadata state. For integration test we need to check that client can push at least two packages one by one and then install it without problems.

PR #121, opened 2021-09-01
Fix routes

### LINKED ISSUE #120: Bug: Fix Gem info routes
Currently requests like GET `/api/v1/gems/{gemname}.json` produce responses for gems, which names don't contain dash `-`. Requests fail if gem name contains dash, like with `file-tail-1.2.0.gem`

milkysoft:
close #120 

Fix routes to return valid gem info.


milkysoft:
@g4s8 , this Pull request is very easy and urgent

src/main/java/com/artipie/gem/http/ApiGetSlice.java
```diff
@@ -53,7 +53,7 @@
     /**
      * Endpoint path pattern.
      */
-    public static final Pattern PATH_PATTERN = Pattern.compile("/api/v1/gems/([\\w]+).(json|yml)");
+    public static final Pattern PATH_PATTERN = Pattern.compile("/api/v1/gems/(.+).(json|yml)");
```
kirill-h5i, line 56: Why to accept any symbol if we need only `-` here?

kirill-h5i (changes requested):
@milkysoft please check my comment

g4s8 (approved):
@milkysoft thanks

Issue #122, opened 2021-09-03
Yaml format


g4s8:
Produce YAML responses for gem info if requested: `GET /api/v1/gems/([\\w]+).yml`

0pdd:
@g4s8 the puzzle [#124](https://github.com/artipie/gem-adapter/issues/124) is still not solved.

0pdd:
@g4s8 the only puzzle [#124](https://github.com/artipie/gem-adapter/issues/124) is solved here.

0pdd:
@g4s8 the puzzle [#137](https://github.com/artipie/gem-adapter/issues/137) is still not solved; solved: [#124](https://github.com/artipie/gem-adapter/issues/124).

PR #123, opened 2021-09-03
feat(yaml): refactored gem info spec format

g4s8:
Changed Gem info spec format inferface to accept not only strings
but hierarchical tree structures. Updated `Gem` to return meta info
interface. Added `JsonManifestFormat` to print structure in JSON.

Ticket: #122


olenagerasimova (approved):
@g4s8 approving, though I think it would be good to add unit tests for new classes. May be, you can add corresponding todos and delegate the jobs to Alexey? If you do not like the idea, just ignore my suggestion.

g4s8:
> @g4s8 approving, though I think it would be good to add unit tests for new classes. May be, you can add corresponding todos and delegate the jobs to Alexey? If you do not like the idea, just ignore my suggestion.

@olenagerasimova done, please check

PR #125, opened 2021-09-06
Add dependencies route

### LINKED ISSUE #97: Gem dependencies API

milkysoft:
close #97


g4s8:
@milkysoft we already have #106 PR for this feature, please don't create duplicate PRs, just push your changes to #106 to preserve the history of changes

PR #126, opened 2021-09-20
refactoring: updated gem sdk, http layer

g4s8:
Added `RubyPlugin` interface to define objects which should be
initialized with Ruby runtime only once.
Refactored some HTTP `Slice`s.


olenagerasimova (approved):
@g4s8 thanks

PR #127, opened 2021-09-24
feat(deps): added core classes for resolving deps

g4s8:
Added `Dependencies` interface to provide API for gem
dependencies. Implemented interface using JRuby, using script
from #125 PR. Added unit test for dependencies.

Ticket: #97


olenagerasimova (approved):
@g4s8 thanks

PR #128, opened 2021-09-27
refactoring: refactored Gem SDK

g4s8:
Removed temporary redundant atomic references,
fixed temp directory removing on future completion,
simplified move operation.


olenagerasimova (approved):
@g4s8 thanks

PR #130, opened 2021-09-29
refactoring(test): refactored gem integration test

g4s8:
Extracted `setUp` and `tearDown` methods to extend test later.
Removed redundant temporary directory mounting, copy data to
container directly.
Removed empty log messages.

Ticket: #97


src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -36,70 +35,100 @@
...
+            .withCommand("tail", "-f", "/dev/null");
+        Testcontainers.exposeHostPorts(port);
+        this.container.start();
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        if (this.server != null) {
+            this.server.stop();
+            this.server.close();
```
olenagerasimova, line 80: As far as I remember, `close()` calls `stop()`, so there is no need to call both methods

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -36,70 +35,100 @@
...
+        }
+        if (this.vertx != null) {
+            this.vertx.close();
+        }
+        if (this.container != null) {
+            this.container.stop();
+            this.container.close();
+        }
```
olenagerasimova, line 88: Same here

src/test/java/com/artipie/gem/GemCliITCase.java
```diff
@@ -36,70 +35,100 @@
...
-        gems.add("file-tail-1.2.0.gem");
         for (final String gem : gems) {
-            final Path target = mount.resolve(gem);
-            try (InputStream is = this.getClass().getResourceAsStream("/".concat(gem));
-                OutputStream os = Files.newOutputStream(target)) {
-                IOUtils.copy(is, os);
-            }
+            this.container.copyFileToContainer(
+                Transferable.of(
+                    IOUtils.toByteArray(this.getClass().getResourceAsStream("/".concat(gem)))
```
olenagerasimova, line 103: Is not it possible to use `TestResourse` from asto and simplify the expression?

olenagerasimova (changes requested):
@g4s8 please check my comments

g4s8:
@olenagerasimova thanks, fixed

olenagerasimova (approved):
@g4s8 thanks

PR #131, opened 2021-09-30
test: integration test for dependencies API

g4s8:
Added integration test with `bundle install` to
check `/api/v1/dependencies` endpoint. Returning
`NOT_IMPLEMENTED` response for this API and disabled
test until implemented.


olenagerasimova (approved):
@g4s8 thanks

PR #132, opened 2021-09-30
feat: gem v1 dependencies API

g4s8:
Added dependencies `Slice` implementation for v1 API,
removed `@Disabled` annotation for test.
Discovered bug in dependencies SDK test, fixed copy
predicate to filter keys by gem names, fixed test.

Close: #97


olenagerasimova (approved):
@g4s8 thanks, looks good to merge

PR #133, opened 2021-10-01
fix: gem info output fields

g4s8:
Reverse-engineered mandatory fields and format in gem API response,
selected them from gem spec ruby object, converted to valid
abstract representation of `MetaInfo`, added unit test,
fixed broken tests, added `@todo` for optional fields.

Fixes: #103


olenagerasimova (approved):
@g4s8 thanks

Issue #124, opened 2021-09-06
JsonMetaFormat.java:34-36: Add tests for this class....


0pdd:
The puzzle `122-d55ec42f` from #122 has to be resolved: 

https://github.com/artipie/gem-adapter/blob/adb502fa1a06940872b7bee26cd254039f54c914/src/main/java/com/artipie/gem/JsonMetaFormat.java#L34-L36

The puzzle was created by @g4s8 on 06-Sep-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #135, opened 2021-10-04
test: unit test for meta json format

g4s8:
Added test case for `JsonMetaFormat`, cover
plain string key-value formats, array formats and
nested objects format.

Closes: #124


olenagerasimova (approved):
@g4s8 thanks

PR #136, opened 2021-10-04
feat: gem meta yaml format

g4s8:
Added `YamlMetaFormat` class to produce YAML outputs for Gem spec.
Added `MetaResponseFormat` enum to produce different responses
on format. Integrated response format to `ApiGetSlice` to return
Gem spec metadata based on request path suffix.

Added `@todo` for tests.

Closes: #122

### DIFF
--- a/src/main/java/com/artipie/gem/http/MetaResponseFormat.java
+++ b/src/main/java/com/artipie/gem/http/MetaResponseFormat.java
@@ -1,0 +1,86 @@
+/*
+ * The MIT License (MIT) Copyright (c) 2020-2021 artipie.com
+ * https://github.com/artipie/artipie/LICENSE.txt
+ */
+package com.artipie.gem.http;
+
+import com.artipie.gem.GemMeta;
+import com.artipie.gem.GemMeta.MetaInfo;
+import com.artipie.gem.JsonMetaFormat;
+import com.artipie.gem.YamlMetaFormat;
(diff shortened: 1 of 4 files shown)


src/main/java/com/artipie/gem/http/MetaResponseFormat.java
```diff
@@ -0,0 +1,80 @@
...
+    YAML {
+        @Override
+        public Response apply(final MetaInfo meta) {
+            final YamlMetaFormat.Yamler yamler = new YamlMetaFormat.Yamler();
+            meta.print(new YamlMetaFormat(yamler));
+            return new RsWithHeaders(
+                new RsWithBody(
+                    StandardRs.OK, yamler.build().toString(), StandardCharsets.UTF_8
+                ),
+                new Header("Content-type", "application/yaml")
```
olenagerasimova, line 54: I'd suggest to use `new ContentType(...)` here and add charset to header value.

olenagerasimova (changes requested):
@g4s8 one suggestion, please, check

src/main/java/com/artipie/gem/http/MetaResponseFormat.java
g4s8: @olenagerasimova thanks, fixed

olenagerasimova (approved):

PR #235, opened 2023-03-07
feat: use new permissions model

### LINKED ISSUE #234: Integrate new permissions model
Let's update asto and http to the current latest versions and use new permissions classes.

olenagerasimova:
closes #234


dgarus (approved):
Thanks!

Issue #261, opened 2023-07-19
Integrate artifact events


olenagerasimova:
Let's update http and add event on artifact upload and repository update.

PR #269, opened 2023-09-18
feat: use standard java queue

olenagerasimova:
part of #261


src/main/java/com/artipie/gem/http/SubmitGemSlice.java
```diff
@@ -78,28 +81,35 @@ public Response response(final String line, final Iterable<Entry<String, String>
...
-                        )
-                    )
-                )
+                        );
+                    } else {
+                        return update.thenAccept(pair -> { });
+                    }
+                }
+            )
                 .thenCompose(none -> this.storage.delete(key))
```
dgarus, line 113: Why do we delete here?

dgarus (approved):
Thanks!
lgtm

